PR #3, opened 2020-05-18
Basic Particle filter implementation

richfitz:
This PR implements the basic particle filter. It is not enough to replace the one in sircovid because I've not yet dealt with the initialisation properly, nor tested models that skip time points


richfitz:
> This looks great, would it be worth adding a regression test with a set seed for the particle filter?

I don't think so at this point because it will just need updating as we tweak the algorithm - I added that on the sircovid case because we were starting with an untested codebase. Typically that sort of test is fragile (as you've seen) and I'm more enthusiastic about trying to get some careful testing here of the underlying statistics

PR #6, opened 2020-05-19
Graceful handling of runs have multiple steps per reporting period

### LINKED ISSUE #4: Confirm working with time skips

richfitz:
This PR adapts the basic particle filter to cope with runs that skip over steps (so if we have 4 steps per day for example).  Do do this we introduce a new function `particle_filter_data` which does basically the same thing as the function of the same name in sircovid and translates our time basis into steps.

Fixes #4


johnlees:
While it is on my mind - how are we dealing with dates in this package? In sircovid the one place I left conversion in was the `particle_filter_data` function. Do we just need to request that data coming in is already converted to a consistent format?

richfitz:
It's not clear - at the moment everything must be an integer but we can put some light date handling into the particle_filter_data function if needed

R/particle_filter_data.R
```diff
@@ -0,0 +1,54 @@
...
+##' mcstate::particle_filter_data(d, "day", 4)
+particle_filter_data <- function(data, time, rate) {
+  assert_is(data, "data.frame")
+  if (!(time %in% names(data))) {
+    stop(sprintf("Did not find column '%s', representing time, in data",
+                 time))
+  }
+
+  t <- data[[time]]
+  assert_integer(t)
```
johnlees, line 40: Do asserts in R give helpful errors? (I know in e.g. C++ they are discouraged as a way of checking parameters)

R/particle_filter.R
```diff
@@ -152,6 +160,7 @@ particle_filter <- R6::R6Class(
 
       self$history <- history
       self$state <- state
+      private$last_model <- model
```
johnlees, line 163: Why is this used rather than `self$model`? Is it possible to modify this outside of the object? If so, why not make it private and add an access function? (Or is that too based upon how I'm used to using objects in C++?)

R/particle_filter_data.R
```diff
@@ -0,0 +1,54 @@
...
+  assert_strictly_increasing(t)
+  assert_integer(rate)
+
+  time_start <- t[-nrow(data)]
```
johnlees, line 44: Took me a minute to understand, maybe add a comment that this adds a lag and lead term/or the start end of discrete bins

R/particle_filter_data.R
johnlees: I now see that they are defined below, so you can ignore this comment

R/utils_assert.R
```diff
@@ -3,4 +3,25 @@ assert_is <- function(x, what, name = deparse(substitute(x))) {
...
                  paste(what, collapse = " / ")), call. = FALSE)
   }
+  invisible(x)
+}
+
+
+assert_integer <- function(x, name = deparse(substitute(x)),
+                           what = "integer") {
+  if (!(is.integer(x))) {
+    usable_as_integer <- is.numeric(x) && (max(abs(round(x) - x)) < 1e-8)
```
johnlees, line 13: Is this a machine epsilon, and why 1e-8?
Looks fine though - does it need to be rounded above too?

R/particle_filter.R
```diff
@@ -109,9 +111,11 @@ particle_filter <- R6::R6Class(
     ##' @param save_history Logical, indicating if the history of all
     ##' particles should be saved
     ##'
+    ##' @param user Optional user parameters to use when creating the model
+    ##'
     ##' @return A single numeric value representing the log-likelihood
     ##' (\code{-Inf} if the model is impossible)
-    run = function(state, n_particles, save_history = FALSE) {
+    run = function(state, n_particles, save_history = FALSE, user = NULL) {
```
johnlees, line 118: Can you overload functions in R? Do you think there would be any use in having a run which also initialises, for cases where the model is only run once? If not, asking for two calls to the object isn't a problem

tests/testthat/test-particle-filter-data.R
```diff
@@ -0,0 +1,32 @@
...
+    particle_filter_data(d, "t", 2.3),
+    "'rate' must be an integer")
+  expect_error(
+    particle_filter_data(d + 0.5, "t", 10),
+    "'t' must be an integer")
+})
+
+
+test_that("particle filter data creates data", {
+  d <- data.frame(day = 0:10, data = 1:11)
```
johnlees, line 22: Might be clearer to make data non-overlapping with day e.g. `data=seq(0.1, 1, 10)`

R/particle_filter_data.R
```diff
@@ -0,0 +1,54 @@
...
+  time_end <- t[-1]
+
+  ret <- data.frame(time_start = time_start,
+                    time_end = time_end,
+                    step_start = time_start * rate,
+                    step_end = time_end * rate,
+                    data[-1, ][names(data) != time])
```
johnlees, line 51: So we have to discard the first observation? Would it be possible to keep this in if there's a constant time step in the data?

R/particle_filter.R
richfitz: The `$model` field is the generator, `$last_model` is the model itself with parameters set in it. I'd hoped to use `$set_user()` but was getting caught between two poor behaviours in odin's handling of the ring buffer. This way is somewhat easier to think about at laest

R/utils_assert.R
richfitz: just pulled in from another package - sqrt(.Machine$double.eps) would be nicer at the cost of computing that every time. Rounding on the way out is a good shout, will implement that

R/particle_filter_data.R
richfitz: I've gone through the sircovid initialisation and updated the code here to reflect that correctly - you were right, we should not discard the first observation but instead infer the previous time. See comment on 97e68ad and the new docs to the function

R/particle_filter.R
richfitz: I'm not sure I follow here, where are two calls to the object?

R/particle_filter_data.R
richfitz: One nice thing about the way that these assertions are implemented is that you get quite nice user-actionable feedback :)

PR #9, opened 2020-06-16
Add a simple grid search, with forecasting and sampling

johnlees:
NB: One of the particle filter tests is problematic - giving a segfault due to something to do with the pointer.

Still to do:
- [ ] Come up with a better test than R^2
- [ ] See if other partitions (other than S, I, R - not actually sure what final two are) need checking
- [ ] Test on between-particle variance vs between-parameter variance


johnlees:
Also, I do not think it is worth putting in all of the plotting, summary or >2D for the grid search. I can think about adding some plotting (maybe more like what's in the orderly task) when the MCMC is added, as that will actually get used

johnlees:
Closing in favour of #10

PR #12, opened 2020-06-22
Add MCMC sampler

### LINKED ISSUE #11: Add MCMC

johnlees:
The MCMC from sircovid, slightly tidied up, less specific (no special handling of beta or dates).
Still needed:
- [x] Some of the old MCMC tests checking proposals
- [x] Improved test coverage

Closes #11


johnlees:
I have left in the purrr call to allow chains to be split across threads. In practise, I think this would need the `long_jump()` method interfaced in dust so that it can be called for each chain thread (then particle threads can safely call jump inside this). We could just take it out for now, if this amount of parallelism is unlikely.

R/forecast.R
```diff
@@ -80,3 +115,42 @@ run_and_forecast <- function(model_params, filter, index, n_particles,
   }
   trajectories
 }
+
+##' @export
+plot.mcstate_forecast <- function(x, ..., what = "1", data = NULL,
+                                  ylab = NULL, title = NULL, col = "grey80") {
+  partition_index <- as.integer(what)
```
richfitz, line 122: Larger and more potentially controversial change - in order to make the plotting testable, can you factor this into all the bits that *would* be passed into the underlying plotting functions, then have another small basically logic-free function for doing the plotting. Created data here might look like:

```
list(particles, title, xlab, ylab, data)
```

(as a named list of course)

richfitz: (The alternative is to get into plot diffing...)

R/particle_filter.R
```diff
@@ -171,7 +171,7 @@ particle_filter <- R6::R6Class(
       }
 
       log_likelihood <- 0
-      prev_state <- model$state()
+      prev_state <- state
       for (t in seq_len(private$n_steps)) {
         model$run(steps[t, 2])
         state <- model$state()
```
richfitz, line 177: At some point we really need to hold something that indicates what state we'd want out at this point - need to decide if that belongs in dust or in mcstate

R/pmcmc.R
```diff
@@ -0,0 +1,609 @@
...
+    chains_coda <- lapply(chains, function(x) {
+        traces <- x$results
+      coda::as.mcmc(traces[, vars$range$name])
+    })
+
+    rhat <- tryCatch(expr = {
+      x <- coda::gelman.diag(chains_coda)
+      x
+    }, error = function(e) {
+      print("unable to calculate rhat")
```
richfitz, line 108: print is the worst named function in R - please use message. Also this error handler should probably return `NA_real_` or `NULL` as it is currently returning the string!

R/pmcmc.R
```diff
@@ -0,0 +1,609 @@
...
+      })
```
richfitz, line 109:
```suggestion
    rhat <- tryCatch(
      coda::gelman.diag(chains_coda),
      error = function(e) NA_real_)
```

☝️ My inclination would be to do this, and not print to screen the message about rhat not working. I'd be inclined to do it in free function for ease of testing though:

```
gelman_diag <- function(chains) {
  tryCatch(
    coda::gelman.diag(chains_coda),
    error = function(e) NA_real_)
}
```

Note that things like:

```r
x <- f(...)
x
```

while common, are redundant

R/pmcmc.R
```diff
@@ -0,0 +1,609 @@
...
+      nrow(proposal_kernel) == length(par_names) &&
+      ncol(proposal_kernel) == length(par_names))) {
+    stop(paste0("proposal_kernel must be a matrix or vector with names ",
+                "corresponding to the parameters being sampled"))
+  }
+
+  # For the chains to be independent on threads, dust will
+  # need to support a long_jump call
+  # (for now, could use seed = seed + thread_idx)
+  chains <- furrr::future_pmap(
```
richfitz, line 82: I am really wondering if we should move the parallel bits into sircovid and just keep the raw mcmc here. Future takes care of much of the work for us, which is nice, but it also dies in irritating ways...

R/pmcmc.R
```diff
@@ -0,0 +1,609 @@
...
+
+## create function to reflect proposal boundaries at pars_min and pars_max
+# this ensures the proposal is symetrical and we can simplify the MH step
+reflect_proposal <- function(x, floor, cap) {
+  interval <- cap - floor
+  abs((x + interval - floor) %% (2 * interval) - interval) + floor
+}
+
+# Check input data.frame
+mcmc_validate_range <- function(range) {
```
richfitz, line 284: Is it worth trying to reduce the overlap here with the same function in grid? They look like they can share a common core

R/pmcmc.R
```diff
@@ -0,0 +1,609 @@
...
+
+  if (any(range$init < range$min | range$init > range$max)) {
+    stop("initial parameters are outside of specified range")
+  }
+
+  list(range = range,
+       index = index)
+}
+
+##' @title create a master chain from a pmcmc_list object
```
richfitz, line 332: What is a master chain? As this is an exported function can it get defined here?

R/pmcmc.R
```diff
@@ -0,0 +1,609 @@
...
+  cols <- viridis::cividis(nrow(traces))
+  cols <- cols[order(order(x$results$log_likelihood))]
+
+  par(bty = "n",
+      mfcol = c(n_pars, n_pars + 1L),
+      mar = c(2.5, 2.5, 2, 1.5),
+      mgp = c(1.5, 0.5, 0),
+      oma = c(1, 1, 1, 1))
+
+  for (i in seq_len(n_pars)) {
```
richfitz, line 432: You might be interested in `pairs`

R/pmcmc.R
```diff
@@ -0,0 +1,609 @@
...
+
+  par(bty = "n",
+      mfcol = c(n_pars, n_pars + 1L),
+      mar = c(2.5, 2.5, 1.5, 0),
+      mgp = c(1.5, 0.5, 0),
+      oma = c(1, 1, 1, 1))
+
+  for (i in seq_len(n_pars)) {
+    for (j in seq_len(n_pars)) {
+      if (i == j) {
```
richfitz, line 529: Also `pairs`

R/pmcmc.R
johnlees: And harder to debug. I am going to remove it, and maybe leave a note about its possible use here

R/pmcmc.R
johnlees: Yes, especially as I started by copying that function

R/forecast.R
johnlees: Leaving for now, raised #15 to track

Issue #41, opened 2020-08-13
Add support for prediction based on a particle filter object?


richfitz:
Removed in 2967912ccf09dae111531a2c56cad6e721d5aa4a but perhaps useful in future? Do we ever need this?

johnlees:
This is when you 'extend' a few example trajectories?

My memory is that this was more needed with the grid search, where you had to first sample the trajectories based on the posterior prob. The trajectories from the MCMC are already a sample of the posterior prob

richfitz:
yeah, I confirmed with @edknock and @lwhittles that it was not actually needed, so rather than continuing to patch it I've ripped it out for now

PR #43, opened 2020-08-13
i37: add forecasting support

### LINKED ISSUE #24: Can't run predict multiple times
Need to reset the state (and time) after prediction back to what it was *before* the prediction

### LINKED ISSUE #37: Redo forecasting
Discussing on Friday. Issues

* nowcasting (simulate consistent with the data)
* forecasting (do not actually run the particle filter)

So we need to make it easy for the MCMC to return the end-state and then we can simulate from it

richfitz:
Adds support for running predictions from a pmcmc run. Predictions will be run in parallel thanks to changes introduced in dust (via https://github.com/mrc-ide/dust/pull/86). 

Other changes of note:

* Some light tidying of the particle filter internals (more to come but the changes here are the minimal set for this PR)
* `particle_filter_data` is now compulsory for use with particle_filter, as this brings in the `time` and `rate` attributes that will use elsewhere
* There is a new `$predict_info()` method that I do not love the name of. It returns information required for running predictions.
* The format of trajectories has changed and they now include time information alongside the state

**NOTE**: This removes support for prediction directly from a particle filter, as that is not needed by us and the interface is very awkward.

Fixes #37 
Closes #24 (by obsoleting it, see #41)


R/particle_filter.R
```diff
@@ -360,85 +365,20 @@ particle_filter <- R6::R6Class(
...
-      if (append) {
-        res <- dde_abind(self$history(), res)
-      }
-
-      res
+    ##' @description
+    ##' Return a list of information that would be used to restart a
+    ##' simulation (in addition to state). This exists to support internal
+    ##' functions and should not need to be called by end-users.
+    predict_info = function() {
```
johnlees, line 372: How about `get_end_state()`, `get_predict_info()` or `get_simulation_state()`

R/particle_filter.R
```diff
@@ -360,85 +365,20 @@ particle_filter <- R6::R6Class(
...
+      ## TODO: this name is terrible.
+      list(n_threads = private$last_model$n_threads(),
```
johnlees, line 374: Add RNG state? (even if we can't load it back in just yet)

R/pmcmc.R
```diff
@@ -119,20 +119,43 @@ pmcmc <- function(pars, filter, n_steps, save_state = FALSE,
...
 
-  state <- trajectories <- NULL
+  predict <- state <- trajectories <- NULL
   if (save_state) {
     state <- t(list_to_matrix(history_state$get()))
+
+    ## Extract the transformation function from the pars object as
+    ## it's all we want; this is a bit of a hack but it works for now
+    ## at least.
+    transform <- pars[[".__enclos_env__"]]$private$transform
```
johnlees, line 134: Haven't seen this before, just double-checking it's a safe way to get this? You can't alter the environment in some way that breaks it?

johnlees: Presumably this isn't possible within `predict_info()`?

R/pmcmc.R
```diff
@@ -119,20 +119,43 @@ pmcmc <- function(pars, filter, n_steps, save_state = FALSE,
...
+                    n_threads = info$n_threads,
+                    index = info$index,
+                    rate = info$rate,
+                    step = last(info$step))
   }
   if (save_trajectories) {
-    trajectories <- list_to_array(history_trajectories$get())
+    info <- filter$predict_info()
+    trajectories_state <-
+      aperm(list_to_array(history_trajectories$get()), c(1, 3, 2))
```
johnlees, line 153: Why 1, 3, 2? Add a comment to this line perhaps

R/pmcmc_tools.R
```diff
@@ -0,0 +1,49 @@
...
+##' @param object Results of running [pmcmc()]
+##'
+##' @param burnin Optional integer number of iterations to discard as
+##'   "burn-in". If given then samples `1:burnin` will be excluded
+##'   from your results. Remember that the first sample represents the
+##'   starting point of the chain. It is an error if this is not a
+##'   positive integer or is greater than the number of samples.
+##'
+##' @param thin Optional integer thinning factor. If given, then every
+##'   `thin`'th sample is retained (e.g., if `thin` is 10 then we keep
```
johnlees, line 15: thin'th – love it

R/predict.R
```diff
@@ -0,0 +1,63 @@
...
+##'   first value must be the final value run in your simulation. An
+##'   error will be thrown if you get this value wrong, look in
+##'   `object$predict$step` (or the error message) for the correct
+##'   value.
+##'
+##' @param n_threads The number of threads used in the simulation. If
+##'   not given, we default to the value used in the particle filter
+##'   that was used in the pmcmc.
+##'
+##' @param seed The random number seed - a positive integer.
```
johnlees, line 21: Note that this will be taken out?

R/pmcmc.R
```diff
@@ -119,20 +119,43 @@ pmcmc <- function(pars, filter, n_steps, save_state = FALSE,
...
 
-  pars <- set_colnames(list_to_matrix(history_pars$get()),
-                       names(curr_pars))
+  pars_matrix <- set_colnames(list_to_matrix(history_pars$get()),
+                              names(curr_pars))
   probabilities <- set_colnames(list_to_matrix(history_probabilities$get()),
                        c("log_prior", "log_likelihood", "log_posterior"))
 
-  state <- trajectories <- NULL
+  predict <- state <- trajectories <- NULL
```
johnlees, line 127: Thoughts about a default of `save_state = TRUE`? If you forget to set it that's going to be annoying and a loss of time, whereas the overhead would be relatively small

tests/testthat/test-pmcmc-tools.R
```diff
@@ -0,0 +1,71 @@
...
+  i <- seq(11, 31, by = 4)
+  res <- pmcmc_thin(results, 10, 4)
+  expect_identical(res$pars, results$pars[i, ])
+  expect_identical(res$probabilities, results$probabilities[i, ])
+  expect_identical(res$state, results$state[, i])
+  expect_identical(res$trajectories$state, results$trajectories$state[, i, ])
+})
+
+
+test_that("can discard whole chain", {
```
johnlees, line 42: I feel like this should error

R/particle_filter.R
richfitz: It's the `predict_info` part of the name I don't like! Generally I am not a fan of `get_` functions. What are you seeing as the difference between end_state and simulation_state?

R/particle_filter.R
```diff
@@ -63,11 +63,13 @@ particle_filter <- R6::R6Class(
   cloneable = FALSE,
 
   private = list(
+    ## Control over the data
     data = NULL,
+    data_split = NULL,
```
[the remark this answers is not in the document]
richfitz: This is the data, split by rows.  So a table

```
a | b
1 | 2
3 | 4
5 | 6
```

becomes a list

```
{a: 1, b: 2}
{a: 3, b: 4}
{a: 5, b: 6}
```

this was already done, but previously we discarded the data itself - the change here is to support retaining the original data as well as the transformed version

R/particle_filter.R
richfitz: I would prefer to add it when we're ready to use it, rather than speculatively add what we might need - it adds a nontrivial size if we get all particles's states, and if we take only the first it supposes how we will implement the feature. And if we want to advance the state at the same time we reference it (which means subsequent calls would be different) it should be done here.

I do understand the temptation as a previous version of the function did have the rng state :grin:

R/pmcmc.R
richfitz: This relies on an implementation detail in R6, rather than an R detail, and is safe here insofar as R6 will not remove it (https://github.com/r-lib/R6/issues/122, and I think that is depended on fairly widely) and we control the parameters class.

I think that we'd be better off working out how to handle method-like fields in the transform object, or exposing the private field as a readonly public one.

R/pmcmc.R
richfitz: comment added in f57c8c1 (this is an array permutation, a generalisation of a transpose)

R/pmcmc_tools.R
richfitz: pity those with a lisp

R/predict.R
richfitz: done in 0237bcd

R/pmcmc.R
richfitz: yeah, that's a good argument. Updated in b8b8e29, along with a comment about the consequences

tests/testthat/test-pmcmc-tools.R
richfitz: I was on the fence on this, will update

R/pmcmc.R
richfitz: > Presumably this isn't possible within predict_info()?

`predict_info` only sees elements of the particle filter, but we need the transform function, which comes from the `pmcmc_parameters` object

R/pmcmc_tools.R
richfitz: (update here is just that I noticed I'd prematurely started using markdown before we support it)

richfitz:
> Main question: is forecasting without running the MCMC going to be possible at some point? So you can just load values in and run (as is being done in the RWC scenarios). Or is that outside of mcstate, and to be done directly with dust instead?

Yeah, this is a good point; we'll add support to mcstate to make this easier but I need to understand what the interface needs to look like - Lilith has started sketching this out. The underlying problems to be solved are similar though: gather the data that dust needs and fire it off in one go

R/particle_filter.R
johnlees: My thought was starting with the verb `predict` makes it sounds like you will run a prediction. What about `extract_` or similar, or is that the same as `get_`. `end_state` probably isn't great, `simulation_state` is better.

johnlees:
> > Main question: is forecasting without running the MCMC going to be possible at some point? So you can just load values in and run (as is being done in the RWC scenarios). Or is that outside of mcstate, and to be done directly with dust instead?
> 
> Yeah, this is a good point; we'll add support to mcstate to make this easier but I need to understand what the interface needs to look like - Lilith has started sketching this out. The underlying problems to be solved are similar though: gather the data that dust needs and fire it off in one go

Noted here https://github.com/mrc-ide/mcstate/issues/44

PR #47, opened 2020-08-14
i46: Use new flexible dust seeding

### LINKED ISSUE #46: Seeding for simulate should be based on the current model state

richfitz:
This PR makes mcstate's seeding more flexible, using the new approach in dust. The default value for seed is `NULL` and it's possible to continue a stream of random numbers in `dust_simulate` using the final state of the RNG which is saved.

* should we save just the first 32 bytes of state into the object? This will be much smaller and also faster to jump. The state is unlikely to compress well on serialisation so for 1k particles we will have 32Kb of data.
* we advance the state of the rng at the save point which guarantees the states are decoupled. But would it make more sense to jump the state of saved state? This means we're not really continuing but are definitely uncorrelated and might simplify use in the parallel case

Depends on mrc-ide/dust#88 and should not be merged until that is (and the travis branch pin removed).

Fixes #46

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ addons:
     packages:
       - libnode-dev
 r_github_packages:
-  - mrc-ide/dust
+  - mrc-ide/dust@i87-rng-seed
 r_packages:
   - covr
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ addons:
     packages:
       - libnode-dev
 r_github_packages:
-  - mrc-ide/dust@i87-rng-seed
+  - mrc-ide/dust
 r_packages:
   - covr
 after_success:
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mcstate
 Title: Monte Carlo Methods for State Space Models
-Version: 0.2.1
+Version: 0.2.2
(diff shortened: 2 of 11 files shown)


.travis.yml
```diff
@@ -6,7 +6,7 @@ addons:
     packages:
       - libnode-dev
 r_github_packages:
-  - mrc-ide/dust
+  - mrc-ide/dust@i87-rng-seed
```
johnlees, line 9: NB: Merge https://github.com/mrc-ide/dust/pull/88 first, then remove this before merging here

richfitz:
> Agree that saving the first part of state would be best (thinking about moving to ~1M particles if serialisation is slow). Loading this in versus calling jump many times is not going to be very different in terms of time anyway.

OK, I'll look at implementing this and I will see how that looks

> I don't understand this distinction I'm afraid - what is the difference you mean between jump and advance?

Sorry "advance" just trying to abstract in my head the idea of jumping the sequence without being too tied down to the xoshiro terminology - part of the motivation there was just to avoid saying "jump" because it's really a "long_jump" and that's just kind of weird to use as an argument!

To expand on the point. What we do in this PR is:

* run the pmcmc
* save the state from the dust object at that point
* advance the dust object's mcmc state by doing a long_jump

The rationalle there is that if you were handing off from the pmcmc to the prediction that would make the stream just continue as before. However, if you want to simulations multiple times that leads to the weirdness as documented

Issue #51, opened 2020-08-17
Parallel chains by using a multiparameter dust object


richfitz:
Should be easy once https://github.com/mrc-ide/dust/issues/92 is implemented

richfitz:
I think this will be less efficient than the current worker method

PR #59, opened 2020-08-19
i58: multiple chains

### LINKED ISSUE #58: Basic support for multiple chains

richfitz:
Simple support for multiple chains.

As used in the ONS task, seems generally useful.  We'll likely need a slightly different interface to support #51 though.

Fixes #58

### DIFF
--- a/R/pmcmc.R
+++ b/R/pmcmc.R
@@ -47,6 +47,11 @@
 ##' @param progress Logical, indicating if a progress bar should be
 ##'   displayed, using [`progress::progress_bar`].
 ##'
+##' @param n_chains Optional integer, indicating the number of chains
+##'   to run. If more than one then we run a series of chains and
+##'   merge them with [pmcmc_combine()]. Chains are run in series,
+##'   with the same filter.
+##'
 ##' @return A `mcstate_pmcmc` object containing `pars`
 ##'   (sampled parameters) and `probabilities` (log prior, log
 ##'   likelihood and log posterior values for these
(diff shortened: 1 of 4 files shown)


R/pmcmc.R
```diff
@@ -61,13 +66,32 @@
 ##'
 ##' @export
 pmcmc <- function(pars, filter, n_steps, save_state = TRUE,
-                  save_trajectories = FALSE, progress = FALSE) {
+                  save_trajectories = FALSE, progress = FALSE,
```
johnlees, line 69: Add the burn-in option here? Or note that it can be thinned/sampled from later.
And also just to check – can you still remove burn in after the chains have been combined, or does it look like one big chain?

R/pmcmc.R
```diff
@@ -61,13 +66,32 @@
...
+        message(sprintf("Running chain %d / %d", i, n_chains))
+      }
+      samples[[i]] <-
+        pmcmc1(pars, filter, n_steps, save_state, save_trajectories, progress)
+    }
+    pmcmc_combine(samples = samples)
+  }
+}
 
+pmcmc1 <- function(pars, filter, n_steps, save_state, save_trajectories,
```
johnlees, line 93: `pmcmc1` for some reason doesn't sit right with me. `pmcmc_single_chain` or `pmcmc_chain` feel more descriptive. Definitely very subjective though, so feel free to keep if you prefer the current name

johnlees (changes requested):

R/pmcmc.R
richfitz: burnin/thin is purposefully done later, and it works as you'd want on a combined chain (I think you reviewed this the other day!)

R/pmcmc.R
richfitz: sure, good call. It's an internal function but I do not feel strongly at all. pmcmc_single_chain is good

R/pmcmc.R
johnlees: haha, sorry. I'd forgotten whether they get put into a single row or into columns - but yes I do recall that you keep the number of chains

johnlees (approved):

PR #76, opened 2020-09-21
Add vignette for odin.dust, dust and mcstate

### LINKED ISSUE #22: Add vignette

johnlees:
This attempts to document the whole workflow, with some slightly more complex examples than previously included.

Questions:

- Anything that would benefit from being added?
- Any objections to documenting all three packages working together here?
- Is the use of the cache for the MCMC runs in the vignette correct?

Closes #22


johnlees:
Not passing on Ubuntu 20.04 (devel):
```
ERROR: dependencies ‘curl’, ‘httr’, ‘oai’ are not available for package ‘socialmixr’
```
but fine on release. A CRAN error?

johnlees:
html here for convenience:
[vignette.zip](https://github.com/mrc-ide/mcstate/files/5261275/vignette.zip)

DESCRIPTION
```diff
@@ -28,8 +28,17 @@ Suggests:
     mockery,
     mvtnorm,
     testthat,
-    withr
```
richfitz, line 31: Can you please sort these alphabetically?

vignettes/deaths.txt
```diff
@@ -0,0 +1,36 @@
+## Definition of the time-step and output as "time"
```
richfitz, line 1: Any reason that this, and other odin models, are not a .R file? If it's just the linting drama that's tuneable, though lintr's file-ignore system seems broken to me

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+  comment = "#>"
+)
+```
+
+```{r setup}
+library(odin.dust)
+library(mcstate)
+library(rmarkdown)
+```
+
```
richfitz, line 22: Probably we need some basic introductory stuff here about what the problem is we're trying to solve

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+```{r setup}
```
richfitz, line 17: I know opinions vary, but I'm in favour of not explicitly loading the libraries

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+$$\begin{align*} 
+n_{SI} &\sim B(S, 1 - e^{-\beta \frac{I}{N} \cdot dt)} \\
+n_{IR} &\sim B(I, 1 - e^{-\gamma \cdot dt)}
+\end{align*}$$
+
+## Implementing the SIR model using [`odin.dust`](https://mrc-ide.github.io/odin.dust/)
+
+The above equations can straightforwardly be written out using the odin DSL:
+
+```r
```
richfitz, line 51: This duplicates your model file, which will lead to confusion. Can you adapt the approach in dust to read in and print to screen.  Look for the part that says

````
```{r pkg_dust_r, echo = FALSE, results = "asis"}
r_output(readLines(file.path(path, "R/dust.R")))
```
````

File: vignettes/sir_models.Rmd [hunk 6]
```diff
@@ -0,0 +1,680 @@
...
+I_{t+1} &= I_t + n_{SI} - n_{IR} \\
+R_{t+1} &= R_t - n_{IR}
+\end{align*}$$
+
+where
+$$\begin{align*} 
+n_{SI} &\sim B(S, 1 - e^{-\beta \frac{I}{N} \cdot dt)} \\
+n_{IR} &\sim B(I, 1 - e^{-\gamma \cdot dt)}
+\end{align*}$$
+
```
richfitz, line 46: For the *very first* example, I'd wondered about using one of the builtin examples. So we have already `dust::dust_example("sir")` which implements this model. The reasoning here is that it's less stuff to start getting started with

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+I_ini <- user(10)
+beta <- user(0.2)
+gamma <- user(0.1)
+```
+
+This is converted to a C++ dust model, and compiled into a library in a single step, using [`odin.dust`](https://mrc-ide.github.io/odin.dust/). Save the above code as a file named `sir.R`. File names must not contain special characters.
+
+```{r}
+library(odin.dust)
+gen_sir <- odin_dust("sir.txt")
```
richfitz, line 90: I'd prefer an explicit `odin.dust::odin_dust()` here

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+```
+
+## Adding age structure to the model
```
richfitz, line 93: This feels like it's in the wrong vignette to me

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+## Transform the matrix to the (symetrical) transmission matrix
+## rather than the contact matrix
+transmission <- contact$matrix /
+  rep(contact$demography$population, each = ncol(contact$matrix))
+transmission
+```
+
+This can be given as a parameter to the model by adding the argument `data = list(m = transmission)` when using the `new()` method on the `gen_age` generator.
+
+## Saving a model into a package
```
richfitz, line 201: This is definitely in the wrong vignette - this belongs in odin.dust

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+If you want to distribute your model in an R package, rather than building it locally, you will want to use the `odin_dust_package()` function instead. This will write the transpiled C++ dust code into `src` and its R interface in `R/dust.R`. Package users are not required to regenerate the dust code, and their compiler will build the library when they install the package. You may also wish to add a file `R/zzz.R` to your package `mycoolpackage` with the following lines:
+
+```r
+##' @useDynLib mycoolpackage, .registration = TRUE
+NULL
+```
+
+to help automate the compilation of the model.
+
+# Running the SIR model with [`dust`](https://mrc-ide.github.io/dust/)
```
richfitz, line 212: IMO all the material above here can be dropped and/or moved to different vignettes. I think as much as possible this vignette should focus on the "I want to do inference" problem probably. We can have as many vignettes as needed and IME keeping them fairly focussed, while hard, is necessary to keep them clear

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+
+Now we can use the `new` method on the generator to make `dust` objects. Typically you will work through the `mcstate` package to run the model, but for illustrative purposes we show here that `dust` can be driven directly from R.
+
+[`new()`](https://mrc-ide.github.io/dust/reference/dust.html#method-new) takes the data needed to run the model i.e. a list with any parameters defined as `user` in the odin code above, the value of the initial step $t_0$, and the number of particles, each for now can simply be thought of as an independent stochastic realisation of the model, but in the next step will be used when inferring model parameters.
+
+Additional arguments include the number of threads to parallelise the particles over, and the seed for the random number generator. The seed must be an integer, and using the same seed will ensure reproducible results for all particles. For threads to take effect, `openmp` must be enabled in the compilation step, which you can check by running:
```
richfitz, line 218: Detail like this probably goes later on in a section on peformance?

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+
+```{r}
+sir_model <- gen_sir$new(data = list(dt = 1, S_ini = 1000, I_ini = 10, beta = 0.2, gamma = 0.1),
+                         step = 1,
+                         n_particles = 10L,
+                         n_threads = 4L,
+                         seed = 1L)
+```
+
+Check the initial state is ten particles wide, with four states deep ($t$, $S$, $I$, $R$):
```
richfitz, line 234:
```suggestion
The initial state is ten particles wide, four states deep ($t$, $S$, $I$, $R$):
```

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+```
+
+Run the particles (repeats) forward 10 steps of length $dt$, followed by another 10 steps:
+
+```{r}
+sir_model$run(10)
+sir_model$run(20)
+```
+
+To change the parameters, the model does not need to be recompiled. We simply use the generator object to make a new model. Let's do this and increase the infection rate, and the population size. We will also use a smaller time step to calculate multiple transitions per unit time:
```
richfitz, line 247: I think that the compilation part here is a red-herring (not the detail the user will be expecting). You might write this as

```suggestion
We can change the parameters, say by increasing the infection rate and the population size, by reinitalising the model. We will also use a smaller time step to calculate multiple transitions per unit time:
```

**However** this goes against advice in the dust vignette to use `$reset()` if you want to change parameters, so you might do this instead with that function

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+x <- x[-1, , ]
+
+par(mar = c(4.1, 5.1, 0.5, 0.5), las = 1)
+matplot(time, t(x[1, , ]), type = "l",
+         xlab = "Time", ylab = "Number of individuals",
+         col = "#8c8cd9", lty = 1, ylim = range(x))
+matlines(time, t(x[2, , ]), col = "#cc0044", lty = 1)
+matlines(time, t(x[3, , ]), col = "#999966", lty = 1)
+legend("left", lwd = 1, col = c("#8c8cd9", "#cc0044", "#999966"), legend = c("S", "I", "R"), bty = "n")
+```
```
richfitz, line 279: Consider something above like

```
cols <- c(S = "#8c8cd9", I = "#cc0044", R = "#999966")
```

so that you can do

```
matplot(..., col = cols[["S"]])
matlines(..., col = cols[["I"]])
legend(..., col = cols, legend = name(cols))
```

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+
+Many other methods are available on the dust object to support inference, but typically you will use the `mcstate` package instead of calling these directly. However, if you wish to simulate state space models with set parameters, this can be done entirely using the commands above from `dust`.
+
+# Inferring parameters with [`mcstate`](https://mrc-ide.github.io/odin.dust/)
+
+Typically these models will be used in conjunction with observed data to infer the most likely parameter values, most likely trajectories up to the present (nowcast), and make forecasts using these parameters. The `mcstate` package implements particle filter and pMCMC methods to make all of this this possible from `dust` models.
```
richfitz, line 285: "most likely" looks incorrect to me as we don't offer any ML methods 😛

richfitz: Also pMCMC needs defining, possibly link to the wikipedia page or an article?

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+
+A comparison function is needed to define a likelihood, i.e. the probability of observing this trajectory, given the data. This is combined with a prior to generate a posterior distribution for each parameter, which in the SIR model gives the distribution of likely values for $\beta$ and $\gamma$. 
+
+This is straightforward for a deterministic model, but for a stochastic model looking at a single run can easily bias these estimates. `mcstate` therefore implements a particle filter, also known as sequential Monte Carlo (SMC), which runs multiple trajectories, and at each stage where data is available resamples them in proportion to their likelihood. This produces an unbiased estimate of the likelihood for a given set of parameters.
```
richfitz, line 289: Does a single run "bias" these estimates?

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+
+The anatomy of an mcstate particle filter consists of three components:
+
+- A set of observations to fit the model to, generated using `particle_filter_data()`.
+- A model to fit, which must be a dust generator.
+- A comparison function, which is an R function which calculates the likelihood of the state given the data.
```
richfitz, line 295: given the data _at a particular point in time_

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+         xlab = "Time", ylab = "Number of individuals",
+         col = "#8c8cd9", lty = 1, ylim = range(x))
+matlines(time, t(x[2, , ]), col = "#cc0044", lty = 1)
+matlines(time, t(x[3, , ]), col = "#999966", lty = 1)
+legend("left", lwd = 1, col = c("#8c8cd9", "#cc0044", "#999966"), legend = c("S", "I", "R"), bty = "n")
+```
+
+Many other methods are available on the dust object to support inference, but typically you will use the `mcstate` package instead of calling these directly. However, if you wish to simulate state space models with set parameters, this can be done entirely using the commands above from `dust`.
+
+# Inferring parameters with [`mcstate`](https://mrc-ide.github.io/odin.dust/)
```
richfitz, line 283: Actually *this* might be where I would start the vignette :)

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+
+- A set of observations to fit the model to, generated using `particle_filter_data()`.
+- A model to fit, which must be a dust generator.
+- A comparison function, which is an R function which calculates the likelihood of the state given the data.
+
+## Model data
+
+To demonstrate this, we will attempt to infer the parameters $\beta$ and $\gamma$ using daily case counts. In reality these counts would be measured, but here we generate them from a single run of the model, and adding some constant noise proportional to the epidemic size:
+
+```{r}
```
richfitz, line 301: My suggestion here is that you provide the data as a fait accompli. You can tell the reader that they were generated from the model (or not), but read them from a csv or hide the sausage factory here

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+  round(rnorm(mean = 0, n = length(true_history[3,1,-1]), sd = 0.1*sqrt(true_history[3,1,-1])))
+incidence <- unlist(lapply(incidence, max, 0))
+incidence_data <- data.frame(cases = incidence, day = seq_len(n_steps))
+incidence
+```
+
+This can then be processed with `particle_filter_data()` as follows:
+
+```{r paged.print=TRUE, layout="l-body-outset"}
+library(mcstate)
```
richfitz, line 335: This is the second inclusion of this library, but I'd prefer ::

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+The comparison function will typically be a likelihood, the probability of the simulation run given the data. This can be any function in R which takes the current simulated `state`, the state at the previous time step `prev_state`, a `data` row for the current time point and a list of any parameters `pars` the function needs.
+
+For this model the number of cases can be calculated from the change in susceptibles between time steps (as above). In this comparison function, the number of cases within each time interval is modelled as being Poisson distributed. A small amount of noise is added to prevent zero expectations.
+
+```{r}
+case_compare <- function(state, prev_state, observed, pars = NULL) {
+  if (is.na(observed$cases)) {
+    return(NULL)
+  }
+  if (is.null(pars$compare$exp_noise)) {
```
richfitz, line 355: This if / else is complicating things here and was needed I think in order to help with testing. If you want to demonstrate tuning the parameter I would unconditionally use your else branch, otherwise I'd unconditionally use your if branch

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+        col = "#999966", lty = 1, ylim = range(history))
+matlines(incidence_data$day, t(history[3, , -1]), col = "#8c8cd9", lty = 1)
+matlines(incidence_data$day, t(history[4, , -1]), col = "#cc0044", lty = 1)
+matpoints(incidence_data$day, t(true_history[2:4, , -1]), pch = 19,
+          col = c("#999966", "#8c8cd9", "#cc0044"))
+legend("left", lwd = 1, col = c("#999966", "#8c8cd9", "#cc0044"), 
+       legend = c("S", "I", "R"), bty = "n")
+```
+
+Typically you will not use the `run()` function of the particle filter directly, and instead use the functions below to infer parameters.
```
richfitz, line 427: If the user was developing a new inference method, doing a sensitivity analysis, or manually doing a grid search they might?

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+
+### Using MCMC to infer parameters
+
+Using this, we can do a MCMC using the Metropolis-Hastings algorithm to infer the values of $\beta$ and $\gamma$ from daily case counts. Let's first describe the parameters we wish to infer, giving a minimum, maximum. For $\gamma$ we will also apply a gamma prior with shape 1 and rate 0.2: $\Gamma(1, 0.2)$:
+
+```{r}
+beta <- pmcmc_parameter("beta", 0.2, min = 0)
+gamma <- pmcmc_parameter("gamma", 0.1, min = 0, 
+                         prior = function(p) {
```
richfitz, line 436: it is typical to omit the curly braces here

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+processed_chains <- pmcmc_thin(pmcmc_run, burnin = 500, thin = 2)
+processed_chains
+apply(processed_chains$pars, 2, mean)
+```
+
+### Optimal use of parallelisation
+
+Note that we run four separate chains in serial here. If you have $m$ nodes available, each with $n$ cores, the most efficient way to run this step is to set `nthreads = n` on the filters, and run $m$ independent chains, one on each node setting `n_chains = 1`. These independent results can then be collected using `pmcmc_combine()`:
+
+```{r pmcmc_serial, cache=TRUE}
```
richfitz, line 473: The cache will only impact your local run, and even then with horribly unpredictable results. I'll rig something up to make this nicer once it settles down as CRAN will not allow a 30 min vignette. However, perhaps just not running this would be best. TBH I am a bit confused here

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+  )
+```
+
+We also provide some basic tools for dealing with this output, before running prediction (see below). We can remove burn-in and thin the four chains as follows:
+
+```{r}
+processed_chains <- pmcmc_thin(pmcmc_run, burnin = 500, thin = 2)
+processed_chains
+apply(processed_chains$pars, 2, mean)
+```
```
richfitz, line 467: Can we plot something here? It would be nice to see a basic plot of the parameters, and a prediction

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+
+### Optimal use of parallelisation
+
+Note that we run four separate chains in serial here. If you have $m$ nodes available, each with $n$ cores, the most efficient way to run this step is to set `nthreads = n` on the filters, and run $m$ independent chains, one on each node setting `n_chains = 1`. These independent results can then be collected using `pmcmc_combine()`:
```
richfitz, line 471: I'm confused here about introducing both the within-chain parallelism and the multiple chains together

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+  pmcmc(
+    mcmc_pars,
+    filter,
+    2000,
+    save_state = TRUE,
+    save_trajectories = FALSE,
+    progress = TRUE,
+    n_chains = 1
+  )
+pmcmc_distributed_run <- pmcmc_combine(pmcmc1, pmcmc2, pmcmc3)
```
richfitz, line 504: Why is this called "distributed" if it was run in series?

richfitz: ah, ok the idea is that these might have been run across different computers? This feels too much a detail of our workflow for an introductory vignette and might fit better in a vignette on scaling. It also avoids the very real problem of correctly getting the RNG streams correct here.

richfitz: With all the above, I'd suggest dropping this section for now

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+
+The probability of infection of a susceptible is then indexed by this force of infection:
+
+```r
+p_SI[] <- 1 - exp(-lambda[i] * dt)
+```
+
+Putting this all together, the age structured SIR model is as follows:
+
+```r
```
richfitz, line 119: This is not really used in the above? It looks like it would fit better in the odin vignette to me

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+summary(mcmc_sir)
+plot(mcmc_sir)
+```
+
+
+## Fitting to multiple datastreams
+
+The set up of the particle filter and observation function makes it easy to fit to multiple datastreams within a Bayesian framework. Let us model a disease where a proportion of infected cases $p_{\mathrm{death}} = 0.05$ die rather than recover by adding the following lines to the odin code:
+
+```r
```
richfitz, line 537: again, dump out the real code here

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+```
+
+```{r}
+gen_death <- odin_dust("deaths.txt")
+```
+
+We can simulate some data to fit to, as before. We will also make the case data patchy, making it missing on some days:
+
+```{r layout="l-body-outset", paged.print=TRUE }
+# Set up model
```
richfitz, line 558: again, use some precooked csv?

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+    return(NULL)
+  }
+  if (is.null(pars$compare$exp_noise)) {
+    exp_noise <- 1e6
+  } else {
+    exp_noise <- pars$compare$exp_noise
+  }
+  
+  ll_pois <- function(obs, model, exp_noise) {
+    if (is.na(obs)) {
```
richfitz, line 617: The is.na branch needs explanation

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+```
+
+Which sends the comparison function the indices listed under `run`, in this case the number of susceptibles, and saves any compartments listed under `state` if the history is kept. In the `compare()` function the index of `state` now corresponds to `index$run`. The missing data must be checked for, and contribute 0 to the likelihood if found:
+
+```{r}
+combined_compare <- function(state, prev_state, observed, pars = NULL) {
+  if (is.na(observed$cases) && is.na(observed$deaths)) {
+    return(NULL)
+  }
+  if (is.null(pars$compare$exp_noise)) {
```
richfitz, line 610: Simplify as above

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+    exp_noise <- 1e6
+  } else {
+    exp_noise <- pars$compare$exp_noise
+  }
+  
+  ll_pois <- function(obs, model, exp_noise) {
```
richfitz, line 616: This function might be lifted out to be a free function for ease of explanation - looks like it does not need the captured scope?

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+incidence[missing] <- NA
+combined_data <- particle_filter_data(
+  data.frame(cases = incidence, deaths = deaths, day = seq_len(n_steps)),
+  time = "day",
+  rate = inv_dt
+)
+paged_table(combined_data)
+```
+
+We then need to add this into our comparison function, and in doing so we will also demonstrate the use of the `index` parameter for the particle filter. The only model outputs needed from the comparison function are $S$ and $D$. For models with a very large state space returning only those compartments required by the comparison function can be important for efficiency. An index function can be defined, which operates on dust generators `info()`:
```
richfitz, line 593: This paragraph hides the motivation in the middle

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+                              compare = combined_compare,
+                              index = index,
+                              n_threads = 4L,
+                              seed = 1L)
+filter$run(save_history = FALSE, pars = list(dt = dt))
+```
+
+and this can be used to run the `pmcmc()` method, as above.
+
+## Running predictions
```
richfitz, line 654: Save the reader bouncing around and put it with the simple model above?

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+
+```{r fig.height=5, fig.width=7}
+forecast <- predict(processed_chains, 
+                    steps = seq(400, 800, 4), 
+                    n_threads = 4L, 
+                    prepend_trajectories = TRUE,
+                    seed = processed_chains$predict$seed)
+
+# Sample 10 particles
+mini_forecast <- forecast$state[,sample.int(dim(forecast$state)[2], size = 10),]
```
richfitz, line 666: dim(x)[2] is nrow(x)

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+time <- mini_forecast[1,1,-1]
```
richfitz, line 667: Looks like linting is not on here by default. Can you look over the style and fix the bulk of these please? I can do this if you'd rather not

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+
+matplot(time*dt, t(mini_forecast[2, , -1]), type = "l",
```
richfitz, line 669: same colour treatment as above (and spaces around `*`)

vignettes/sir_models_cache/html/__packages
```diff
@@ -0,0 +1,5 @@
+base
+odin.dust
+mcstate
+socialmixr
+rmarkdown
```
richfitz, line 5: Please delete the cache files and add a suitable pattern to the gitignore

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,680 @@
...
+dim(p_SI) <- N_age
+dim(m) <- c(N_age, N_age)
+dim(s_ij) <- c(N_age, N_age)
+dim(lambda) <- N_age
+```
+
+As before, save the file, and use `odin.dust` to compile the model:
+
+```{r}
+gen_age <- odin_dust("sirage.txt")
```
richfitz, line 174: This appears never to be used? I'd remove this section entirely for now

richfitz (changes requested):
Sorry, lots of comments through here. I think something that is missing is a landing motivation. Something that says "you're a user and you have time series data that you want to fit a stochastic model to - htf are you going to do that?". The functional interface to mcstate with its index/compare/whatever functions probably needs a bit of high-level guidance too.

Docs are really hard.

I've attached the absolute skeleton of a doc I started working on ages ago. My hope had been to outsource most of the prose to someone else...

[mcstate.txt](https://github.com/mrc-ide/mcstate/files/5278055/mcstate.txt)

vignettes/sir_models.Rmd
johnlees: Moving everything above the mcstate section to odin.dust

vignettes/sir_models.Rmd
johnlees: Yes, starting here

vignettes/sir_models.Rmd
johnlees: Yes this is wrong, I've read a review briefly and understand this a bit better now

vignettes/sir_models.Rmd
johnlees: Excellent idea, this is confusing

vignettes/sir_models.Rmd
johnlees: Moved to another vignette, we can decide whether to keep or remove separately. I think it's an important advantage, so worth having somewhere from the get go. Agree it's too involved for the intro though

vignettes/sir_models.Rmd
johnlees: I've put in a plot function as above, but the density of trajectories may leave something to be desired. `coda` is used to plot mcmc results (as we removed all plotting functions), and prediction is dealt with in a section below

johnlees:
Will look at this more tomorrow, but for now:

- I've moved most of the start to https://github.com/mrc-ide/odin.dust/tree/vignette and will open a PR there when ready
- I've added more of an intro
- I've put the parallelisation information into its own vignette
- I've added some of your earlier vignette into it's own place, as I think it's useful to show that we tested the particle filter implementation correctly

I still need to:

- [ ] Turn on linting
- [ ] Fix the linting errors
- [ ] Check the parallelisation vignette
- [ ] Check the Kalman vignette

johnlees:
@richfitz linting the vignettes is a mystery to me unfortunately. If I run `lintr::lint_package()` they are ignored, same with `lint_dir`. You can run `lint` on the Rmd files directly and it works, but I don't know how to do this just from the `.lintr` file.

https://github.com/jimhester/lintr/blob/master/NEWS.md suggests this should run by default as of v2.0.1, so I'm not sure what's going on

vignettes/sir_models.Rmd
johnlees: @richfitz linting the vignettes is a mystery to me unfortunately. If I run lintr::lint_package() they are ignored, same with lint_dir. You can run lint on the Rmd files directly and it works, but I don't know how to do this just from the .lintr file.

https://github.com/jimhester/lintr/blob/master/NEWS.md suggests this should run by default as of v2.0.1, so I'm not sure what's going on

johnlees: Also I am unable to/incorrectly turning off linting of the odin code in CodeFactor

johnlees:
Updated vignettes HTML: 
[mcstate_vignettes_v2.zip](https://github.com/mrc-ide/mcstate/files/5291589/mcstate_vignettes_v2.zip)

.lintr
```diff
@@ -4,4 +4,4 @@ linters: with_defaults(
     todo_comment_linter = NULL,
     cyclocomp_linter = NULL
     )
-exclusions: list("tests/testthat.R")
+exclusions: list("tests/testthat.R", "vignettes/deaths.R")
```
richfitz, line 7: lintr exclusions seem totally broken, but we can exclude in the codefactor interface

vignettes/sir_incidence_data.txt
```diff
@@ -0,0 +1 @@
+3,2,2,1,1,4,2,5,6,6,8,4,5,7,8,13,14,13,16,14,16,25,17,17,21,18,16,15,16,19,12,11,19,11,22,13,20,15,19,17,22,23,14,8,13,10,20,15,9,14,6,8,5,9,7,1,6,7,2,7,8,5,4,7,2,8,6,3,1,2,5,2,0,3,2,1,1,1,2,3,1,3,0,2,1,2,0,0,2,0,4,1,1,2,0,2,1,0,1,0
```
richfitz, line 1: As discussed on teams:

* can this please be saved as a .csv with both columns
* can the script to generate it be put into the `scripts` directory; write as a standalone .R file that runs assuming that it is run from the package root. You can use `pkgload::load_all()` to load internal functions if needed

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,399 @@
...
+- A model to fit, which must be a dust generator, either `dust::dust()` or `odin.dust::odin_dust()`.
+- A comparison function, which is an R function which calculates the likelihood of the state given the data at one time point.
+
+## Model data
+
+To demonstrate this, we will attempt to infer the parameters $\beta$ and $\gamma$ using daily case counts. In reality these counts would be measured, but here we have generated them from a single run of the model with `seed = 2L`, and added some constant noise proportional to the epidemic size:
+
+```{r}
+# Set up model
+incidence <- t(read.table("sir_incidence_data.txt", sep=",", header = F))
```
richfitz, line 82: See the above comment about where this data should come from

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,399 @@
...
+incidence <- data.frame(cases = incidence[, 1], day = seq_len(100), row.names = NULL)
+```
+
+This can then be processed with `mcstate::particle_filter_data()` as follows, taking four update steps between every observation (generally decreasing $dt$ will improve resolution, especially if rates of change are high, at the expense of extra computation time):
+
+```{r paged.print=TRUE, layout="l-body-outset", fig.height=5, fig.width=7}
+dt <- 0.25
+sir_data <- mcstate::particle_filter_data(data = incidence,
+                                          time = "day",
+                                          rate = 1/dt)
```
richfitz, line 92: This use of `dt` begs for more explanation

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,399 @@
...
+n_particles <- 100
+filter <- mcstate::particle_filter$new(data = sir_data,
+                                       model = gen_sir,
+                                       n_particles = n_particles,
+                                       compare = case_compare,
+                                       n_threads = 4L,
+                                       seed = 1L)
+```
+
+We can now run the particle filter forward, which will run and resample 100 trajectories, and return the final likelihood. It is important to set the correct time-step $dt$ here, as we are using 0.25, rather than the default value of 1 defined in the odin model. We will also save the history, which allows us to plot the particle trajectories, as well as use the likelihood:
```
richfitz, line 153: `$dt` should become `dt` in backticks for consistency

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,399 @@
...
+```
+
+## Running predictions
+
+Going back to the simple SIR model above, we can easily continue running the fitted model forward in time using the posterior estimated by `pmcmc()` to create a forecast past the end of the data:
+
+```{r fig.height=5, fig.width=7}
+forecast <- predict(processed_chains,
+                    steps = seq(400, 800, 4),
+                    n_threads = 4L,
```
richfitz, line 295: throughout this vignette, please remove all uses of the `n_threads` argument.

The reasons for doing this are that we need to not use more than 2 threads on CRAN, but also it complicates writing the parallel vignette

vignettes/sir_models.Rmd
```diff
@@ -0,0 +1,399 @@
...
+gen_death <- odin.dust::odin_dust("deaths.R")
+```
+
+We can simulate some data to fit to, as before. We have also made the case data patchy, so it is missing on some days:
+
+```{r layout="l-body-outset", paged.print=TRUE }
+# Set up model
+dt <- 0.25
+death_data <- read.table("death_data.csv", header = TRUE, sep = ",")
+true_history_deaths <- readRDS("deaths_true_history.rds")
```
richfitz, line 329: All these rds files need saving in the same script as the .csv - can you save them with `saveRDS(..., version = 2)` so that we use a version of rds that works with older R?

richfitz (changes requested):
Minor tweaks, then I'll return the favour with a parallel vignette

richfitz (approved):
As discussed, the parallel vignette will get reworked at some point, but it's good to have a placeholder at least

Issue #75, opened 2020-09-14
Deal with zero probabilities


johnlees:
For some models it is possible to get zero probability/-Inf log prob for all particles. For example, a proposal which sets a binomial probability to 0 or 1 will give this for any state where any observation is not 0/n respectively. This causes issues in the initialisation.

While it is possible to deal with this in the model, or perhaps by setting careful min/max parameter values to avoid this, it would be useful to deal with it here. Ed suggested checking for this, and rerunning the particles (up to a max number of tries). For the MCMC, it shouldn't cause a problem as the proposal won't be accepted, but at the moment it causes errors in the particle filter.

PR #79, opened 2020-10-08
Particle filter partial run (#78)

### LINKED ISSUE #78: Refactor particle filter to allow partial runs
To support https://github.com/mrc-ide/mcstate/issues/13 we need a way of running the particle filter more flexibly. Doing this should not affect the current runs, and would be worth doing somewhat separately to limit the size of the SMC2 implementation PR.

richfitz:
This PR refactors the main loop of the particle filter so that it can be partially run. It should have no effect on running any existing code, but sets us up with the facilities to run SMC2.

It adds minimal user-visible features, with the exception of removing the unused `unique_particles` field from the particle filter class; I think that this was there to support the SMC2 algorithm in any case, but it's not needed. Other than that, note that all tests are additions.

The work in the parameter validation splits out the components that will be needed for the SMC2 parameters, without change.

With this PR in place, the hope is that the algorithm for SMC2 will more clearly come through via the diff

Fixes #78 
~Requires https://github.com/mrc-ide/dust/pull/104 (and the branch pointer should be updated before merge)~


richfitz:
So, updating the branch seems to have broken build due to the branch not being found in dust, but that branch is no longer referenced anywhere. I suspect rerunning tomorrow will be enough...

R/particle_filter.R
```diff
@@ -394,6 +330,156 @@ particle_filter <- R6::R6Class(
...
+    n_particles = NULL,
+    n_threads = NULL,
+    initial = NULL,
+    index = NULL,
+    compare = NULL
+  ),
+
+  public = list(
+    model = NULL,
+    step_current = 0L,
```
johnlees, line 354: Should this be set to `steps[[1L]]`?

R/particle_filter.R
```diff
@@ -394,6 +330,156 @@ particle_filter <- R6::R6Class(
...
+        self$state <- state
+        if (save_history) {
+          self$history$order[, step + 1L] <- seq_len(ncol(state))
+        }
+      } else {
+        weights <- scale_log_weights(log_weights)
+        self$log_likelihood <- self$log_likelihood + weights$average
+        if (self$log_likelihood == -Inf) {
+          self$complete <- TRUE
+          return(-Inf)
```
johnlees, line 442: This seems reasonable. Does this work with `sample` ok? In which case I think this probably also closes #75 (could you add a test for that and it can be closed here - one possibility is zero lambda for Poisson and observed > 0)

R/particle_filter.R
```diff
@@ -394,6 +330,156 @@ particle_filter <- R6::R6Class(
...
+          self$history$order[, step + 1L] <- kappa
+        }
+      }
+
+      self$log_likelihood
+    },
+
+    fork = function(pars) {
+      ## Create a new particle_filter_run object based on our current
+      ## one; same model
```
johnlees, line 461:
```suggestion
      ## Create a new particle_filter_run object based on our current
      ## one; same model, different parameters
```

johnlees (changes requested):
Make sense to separate this out into its own PR, and I like the interface. We'll need some guide on how to use it, but I am sure that will come with #77.

Do we still plan to have a MultiDust object in the C++ at some point? Will this interface easily fit in with that?

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: mcstate
 Title: Monte Carlo Methods for State Space Models
-Version: 0.2.13
+Version: 0.2.14
```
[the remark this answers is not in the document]
richfitz: We don't have a NEWS file but the hope here is that this change is user invisible 😈  - was thinking we should start one for the SMC2 addition...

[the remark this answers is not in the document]
johnlees: Sounds fine!

R/particle_filter.R
richfitz: Two issues here: first, that's not known at the point that this is set, as this is on the generator. Second, this is the index of the steps in `steps`! So the step might start on odin step 100, but that's the first step. So this is an index into `steps`. Ideas for a better name appreciated

R/particle_filter.R
richfitz: This is just the same behaviour as before, so #75 still needs fixing. I think I discussed this with @edknock and sketched a partial solution

R/particle_filter.R
johnlees: `current_step_index`?

DESCRIPTION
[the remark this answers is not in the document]
richfitz: added a news item

R/particle_filter_state.R
```diff
@@ -0,0 +1,220 @@
...
+    history = NULL,
+
+    ##' @field restart_state Full model state at a series of points in
+    ##'   time, if the model was created with non-`NULL` `save_restart`.
+    ##'   This is a 3d array as described in [mcstate::particle_filter]
+    restart_state = NULL,
+
+    ##' @field log_likelihood The log-likelihood so far. This starts at
+    ##'   0 when initialised and accumulates value for each step taken.
+    log_likelihood = NULL,
```
johnlees, line 54: To fix the SMC^2 issue we need a non-accumulating ll, which I think makes sense in this state object. Perhaps `current_log_likelihood`. This will store the value from the current application of compare only.

Does it make sense to add this now so I can use it when I try and apply the fix on the other PR? Or add that all at once?

R/particle_filter_state.R
```diff
@@ -0,0 +1,220 @@
...
+    ##' documented.
+    initialize = function(pars, generator, model, data, data_split, steps,
+                          n_particles, n_threads, initial, index, compare,
+                          seed, save_history, save_restart) {
+      ## NOTE: this will generate a warning when updating docs but
+      ## that's ok; see https://github.com/r-lib/roxygen2/issues/1067
+      if (is.null(model)) {
+        model <- generator$new(pars = pars, step = steps[[1L]],
+                               n_particles = n_particles, n_threads = n_threads,
+                               seed = seed)
```
johnlees, line 67: Is it now possible for us to use a single dust model for all of the particle filters here, by using the index of parameter membership? Or is that too complex to begin with, and we should do that in a future PR?

R/particle_filter_state.R
```diff
@@ -0,0 +1,220 @@
...
+                                       private$pars)
+      }
+
+      if (is.null(log_weights)) {
+        if (save_history) {
+          self$history$order[, step + 1L] <- seq_len(ncol(state))
+        }
+      } else {
+        weights <- scale_log_weights(log_weights)
+        self$log_likelihood <- self$log_likelihood + weights$average
```
johnlees, line 167: If you add log_likelihood_current add here:
```
self$log_likelihood_current <- weights$average
```

R/particle_filter_state.R
```diff
@@ -0,0 +1,220 @@
...
+    ##'
+    ##' @param pars New model parameters
+    fork = function(pars) {
+      seed <- self$model$rng_state()
+      save_history <- !is.null(self$history)
+      ret <- particle_filter_state$new(
+        pars, private$generator, NULL, private$data, private$data_split,
+        private$steps, private$n_particles, private$n_threads,
+        private$initial, private$index, private$compare, seed,
+        save_history, private$save_restart)
```
johnlees, line 207: Similarly, if we use a single dust object, this would also change

johnlees (changes requested):

R/particle_filter_state.R
richfitz: I can do that here, I think; that field is actually redundant atm

R/particle_filter_state.R
richfitz: We should definitely explore that once we get things working. Will be interesting to see when that's faster than running them sequentially or lower-core numbers of particle filters in parallel....

R/particle_filter_state.R
richfitz: Done in 54fc4e9

johnlees (approved):

PR #85, opened 2020-12-21
Add new rerun_every argument to rerun particle filter

richfitz:
This PR adds an argument `rerun_every` to the `pmcmc` function that will rerun the particle filter periodically on the *accepted* point, reducing the chance of stuck chain at the cost of some bias

### DIFF
--- a/R/pmcmc.R
+++ b/R/pmcmc.R
@@ -60,6 +60,14 @@
 ##'   `n_chains` columns to use a different starting point for each
 ##'   chain.
 ##'
+##' @param rerun_every Optional integer giving the frequency at which
+##'   we should rerun the particle filter on the current "accepted"
+##'   state.  The default for this (`Inf`) will never rerun this
+##'   point, but if you set to 100, then every 100 steps we run the
+##'   particle filter on both the proposed *and* previously accepted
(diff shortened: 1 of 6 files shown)


edknock:
I might be wrong, but does this always pull trajectories from the latest run of the particle filter? So in a 'rerun_every' step, you are automatically saving the rerun value, irrespective of whether you accepted the proposal or not? https://github.com/mrc-ide/mcstate/blob/i84-rerun-pf/R/pmcmc.R#L167

R/pmcmc.R
```diff
@@ -137,6 +146,17 @@ pmcmc_single_chain <- function(pars, initial, filter, n_steps,
   for (i in seq_len(n_steps)) {
     tick()
 
+    if (i %% rerun_every == 0) {
+      curr_llik <- filter$run(pars$model(curr_pars), save_trajectories)
+      curr_lpost <- curr_lprior + curr_llik
+      if (save_trajectories) {
```
edknock, line 152: Do we need `particle_idx <- sample.int(n_particles, 1)` in here?

edknock: Above `if (save_trajectories) {`

edknock (approved):
Looks good now

PR #87, opened 2020-12-22
Save state (#86)

### LINKED ISSUE #86: Allow saving of full model state periodically for restarting simulations

richfitz:
Allows the particle filter to save the complete internal state during a run so that it may be restarted.

Allow this to be saved from the pmcmc

Then demonstrate how this can be used to restart the pmcmc

Fixes #86


MJomaba (approved):
Looking good!

PR #99, opened 2021-01-28
Add support for fixing a subset of parameters

### LINKED ISSUE #98: Support for fixing some parameters
After creating a parameters object, should be able to exclude some parameters

richfitz:
We need this to simplify the restart, fixing betas in the past.

Fixes #98


lwhittles (approved):
I have checked this over and it looks great - really elegant solution
My sum total contribution is finding one (1) typo

PR #101, opened 2021-02-04
Separated fixed and varied parameters in objects and pmcmc methods

RaphaelS1:
* Add `pmcmc_varied_parameter` function for creating a varied parameter, which can change between populations. Assumes that `pmcmc_parameter` objects are fixed parameters that are constant between populations.
* Adds associated tests
* Expands `pmcmc_parameters` to detect and auto-tag fixed and varied parameters
* Adds `population` field to `pmcmc_parameters` for naming population
* Updates proposal, summary, prior methods in `pmcmc_parameters`
* Updates `run` method in `pmcmc_state`


R/pmcmc_parameters.R
```diff
@@ -132,50 +134,142 @@ pmcmc_parameters <- R6::R6Class(
...
+            dim = c(dim(proposal), length(populations)))
+          if (!is.null(dimnames(proposal))) {
+            dimnames(arr_proposal) <- c(dimnames(proposal), list(populations))
+          }
+          proposal <- arr_proposal
+          dims <- dim(proposal)
+        }
+
+        # catch wrong dimension sizes
+        if (dims[3] != length(populations) ||
```
richfitz, line 180: perhaps
```
if (!all(dims, c(length(populations), length(parameters), length(parameters)) {
```

R/pmcmc_parameters.R
```diff
@@ -132,50 +134,142 @@ pmcmc_parameters <- R6::R6Class(
...
+            !all(dims[1:2] == length(parameters))) {
+          stop(sprintf(
+            "Expected proposal array with dimensions %d x %d x %d.",
```
richfitz, line 183: Also by this point I am inclined to pull the checking code out into a free function as it's a bit much here (it will be much easier to test as free functions too). See `check_parameters` and `check_save_restart` for possible inspiration

R/pmcmc_parameters.R
```diff
@@ -132,50 +134,142 @@ pmcmc_parameters <- R6::R6Class(
...
+            length(parameters), length(parameters), length(populations)))
         }
+
+        if (!is.null(dimnames(proposal))) {
+          ## At this point we could reorder if that's useful and these
+          ## are setequal.
+          ok <- identical(rownames(proposal), names(parameters)) &&
+            identical(colnames(proposal), names(parameters)) &&
+            identical(dimnames(proposal)[3], populations)
+          if (!ok) {
```
richfitz, line 193: We might treat the names of parameters and populations here separately; depending on how names get added one could reasonably have none, just populations, just pars, both. Where they exist they should be consistent

R/pmcmc_parameters.R
```diff
@@ -132,50 +134,142 @@ pmcmc_parameters <- R6::R6Class(
...
+            stop("Expected dimension names of 'proposal' to match parameters
+            and populations.")
+          }
+        }
+
+         private$proposal <- apply(proposal, 3, rmvnorm_generator)
+         private$min <- sapply(parameters, function(.x) {
```
richfitz, line 200: Please don't use `sapply`

The function itself can be simplified given the checks above
```r
private$min <- vapply(parameters, function(x) rep_len(x, length(populations)),
                      numeric(length(populations)))
```

R/pmcmc_parameters.R
```diff
@@ -132,50 +134,142 @@ pmcmc_parameters <- R6::R6Class(
...
-      }
-      if (!is.null(dimnames(proposal))) {
-        ## At this point we could reorder if that's useful and these
-        ## are setequal.
-        ok <- identical(rownames(proposal), names(parameters)) &&
-          identical(colnames(proposal), names(parameters))
-        if (!ok) {
-          stop("Expected dimension names of 'proposal' to match parmeters")
+        # proposal
+        assert_is(proposal, "array")
```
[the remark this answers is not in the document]
richfitz: Aha, I missed the switch above on parameters being varied; that's perfect (though can you expand the docs for the constructor).

Here, we might do:

```r
if (is.matrix(proposal)) {
  ## usual checks that the proposal is ok, then
  proposal <- array(proposal, c(dim(proposal), length(populations)))
}
```

which would allow a graceful move to use the interface

R/pmcmc_parameters.R
```diff
@@ -132,50 +134,142 @@ pmcmc_parameters <- R6::R6Class(
...
-      private$min <- vnapply(private$parameters, "[[", "min",
-                             USE.NAMES = FALSE)
-      private$max <- vnapply(private$parameters, "[[", "max",
-                             USE.NAMES = FALSE)
     },
 
     ##' @description Return the initial parameter values as a named numeric
-    ##' vector
+    ##' vector or named matrix for multiple populations.
+    # FIXME - COULD CONSIDER ALWAYS RETURNING MATRIX FOR CONSISTENCY
```
richfitz, line 258: The consistency thing here is a real trick. To me this really smells like a different type of object. The biggest issue I have with the design that we're going to have to have all our methods become these big if/else blocks that are essentially determined at construction time.

If we used a different name for this sort of object (`pmcmc_parameters_nested` or whatever) and had that _contain_ `pmcmc_parameters` objects in a list it might work better? It would match the same interface as the `pmcmc_parameters` in function names but with additional dimensions as expected

R/pmcmc_parameters.R
```diff
@@ -306,3 +498,44 @@ check_parameters <- function(parameters, type) {
   names(parameters) <- nms
   parameters
 }
+
+# p <- pmcmc_parameters$new(
```
richfitz, line 502: can you please remove commented code? If you want to turn this into tests or examples that's fine

R/pmcmc_state.R
```diff
@@ -101,18 +101,61 @@ pmcmc_state <- R6::R6Class(
...
-        prop_lpost <- prop_lprior + prop_llik
-
-        if (runif(1) < exp(prop_lpost - private$curr_lpost)) {
-          private$curr_pars <- prop_pars
-          private$curr_lprior <- prop_lprior
-          private$curr_llik <- prop_llik
-          private$curr_lpost <- prop_lpost
-          private$update_history()
-        }
+        # if (inherits(pars, "pmcmc_parameters_shared")) {
```
richfitz, line 104: remove for now please

R/pmcmc_varied_parameter.R
```diff
@@ -0,0 +1,66 @@
...
+        prior <- rep(prior, lng)
+      } else if (length(prior) != lng) {
+        stop(sprintf("Length of 'prior' must be '1' or %s.", lng))
+      }
+      if (any(!vlapply(prior, inherits, what = "function"))) {
+        stop("'prior' should be a list of functions.")
+      }
+    }
+
+    mapply(function(x, y) {
```
richfitz, line 46: This is awkward to read, but not sure on this for now

R/utils.R
```diff
@@ -110,8 +110,25 @@ set_into <- function(x, at, value) {
   x
 }
 
+set_minto <- function(x, atx = NULL, aty = NULL, value) {
```
richfitz, line 113: what is this for? seems unused

richfitz: on a style point, I strongly discourage compulsory arguments after optional ones

R/pmcmc_parameters.R
```diff
@@ -184,28 +278,83 @@ pmcmc_parameters <- R6::R6Class(
...
+        if (is.null(population)) {
+          # FIXME - not convinced if best representation, could consider array
+          # or force 1 pop
+          pops <- lapply(private$populations, function(.x) {
+            data_frame(
+              name = self$names(),
+              min = private$min[rownames(private$min) %in% .x],
+              max = private$max[rownames(private$max) %in% .x],
+              discrete = private$discrete,
+              type = ifelse(self$names() %in% private$varied, "varied", "fixed")
```
richfitz, line 305: Please avoid ifelse, it leads to pain more often than not

richfitz (changes requested):
OK some suggestions here:

* I don't think that this extension to pmcmc_parameters is correct as every method ends up with an if/else block that is determined at construction - so we're looking for a new object type I think. That will simplify things long term I suspect
* There's a few style things here; some will seem arbitrary I suspect :)
* We need a better name than "population"

R/pmcmc_parameters.R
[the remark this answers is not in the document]
RaphaelS1: Also matrices inherit from array so `assert_is` doesn't break if proposal is `matrix` as before

R/pmcmc_parameters.R
```diff
@@ -132,50 +134,142 @@ pmcmc_parameters <- R6::R6Class(
...
+    ##' used, as dust models require this. However, if you need to
     ##' generate derived parameters from those being actively sampled
     ##' you can do arbitrary transformations here.
-    initialize = function(parameters, proposal, transform = NULL) {
-      parameters <- check_parameters(parameters, "pmcmc_parameter")
+    ##'
+    ##' @param populations Specifies the names of the different populations.
+    ##' Ignored if no varied parameters are included.
+    initialize = function(parameters, proposal, transform = NULL,
+                          populations = NULL) {
```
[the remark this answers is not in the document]
RaphaelS1: 'population' here in the general statistical term, i.e. a population to sample from, and then the parameters are estimates of a sample drawn from these. Alternatively 'levels' might make more sense (though could be confused with factor levels).

R/pmcmc_parameters.R
RaphaelS1: My only concern about this is it may expect too much from the user, i.e. they may reasonably ask "once I've created a special parameter object, why do I also need a special set". I see the point with much neater code internally of course and also w.r.t. returning same outputs. 

I guess clear documentation and maybe a specialised vignette might make this clearer, e.g. "use these objects *only when*" and then it's clear they are for specific (not the general) use-case. We'd still need an assertion in `pmcmc_parameters` to check no varied parameters are included

R/pmcmc_parameters.R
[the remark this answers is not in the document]
richfitz: So they do. I have a feeling this changed not that long ago (perhaps for R 4.0.0?)

R/utils.R
RaphaelS1: was originally used to set values into matrices, but you're right is unused, will remove.

R/pmcmc_parameters.R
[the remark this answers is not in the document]
RaphaelS1: as discussed updated in docs, can likely be better worded

R/pmcmc_varied_parameter.R
RaphaelS1: Could change to something like (pseudo):

```R
err <- mapply(x(y), {
    value <- try(x(y), silent = TRUE)
    if (class(value)[1] != "try-error") is.finite(value) else FALSE
})
if (any(err)) stop(infinite or not evaluated)
```

R/pmcmc_parameters.R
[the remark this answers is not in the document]
RaphaelS1: I think so too

RaphaelS1:
@richfitz draft 2 with containers ready, all tests locally passing (haven't checked coverage/lint/docs etc.)
I've left the `fix` method for now as it fundamentally changes each time object changes so I'll add once we're satisfied.

tests/testthat/test-pmcmc-nested-parameters.R
```diff
@@ -0,0 +1,430 @@
...
+  expect_equal(p$model(p$initial()), apply(p$initial(), 1, as.list))
+
+  p <- pmcmc_nested_parameters$new(parameters, proposal_varied, proposal_fixed,
+    transform = function(x) as.list(log(x)))
+
+  expect_equal(p$model(p$initial()),
+                apply(p$initial(), 1, function(x) as.list(log(x))))
+})
+
+test_that("pmcmc_nested_parameters propose", {
```
richfitz, line 336: For new code it's always hard, but as things progress it's nice to break down test blocks until you can write a short sentence about what the test is testing. So something like:

```
test_that("pmcmc_nested_parameters propose changes subset of theta" {
```

and then include the test. If the test blocks then start getting repetitive, farm the setup code into a helper- file

R/pmcmc_nested_parameters.R
```diff
@@ -0,0 +1,488 @@
...
+      }
+    },
+
+    ##' @description Apply the model transformation function to a parameter
+    ##' vector.
+    ##'
+    ##' @param theta a parameter vector in the same order as your
+    ##' parameters were defined in (see `$names()` for that order.
+    model = function(theta) {
+      apply(theta, 1, private$transform)
```
richfitz, line 282: I am not convinced that this will *always* return a list unless `private$transform` does. I think that's fine for now because dust requires a list for both

R/pmcmc_nested_parameters.R
```diff
@@ -0,0 +1,488 @@
...
+    ##' @description Return the initial parameter values as a named numeric
+    ##' vector or named matrix for multiple populations.
+    initial = function() {
+
+      varied <- vapply(private$varied_parameters, function(x) x$initial(),
+        numeric(length(private$param_names$varied)))
+
+      # edge case - 1 varied
+      if (length(private$param_names$varied) == 1) {
+        varied <- t(as.matrix(varied))
```
richfitz, line 184: `t(as.matrix(x))` => `matrix(x, nrow = 1)` which here you can do as

```r
varied <- matrix(varied, dimnames = list(private$param_names$varied, names(varied)))
```

R/pmcmc_nested_parameters.R
```diff
@@ -0,0 +1,488 @@
...
+        rownames(varied) <- private$param_names$varied
+      }
+
+      if (is.null(private$fixed_parameters)) {
```
richfitz, line 188: `length(private$fixed_parameters) == 0` might capture intent better than `is.null`

R/pmcmc_nested_parameters.R
```diff
@@ -0,0 +1,488 @@
...
+        t(varied)
+      } else {
+        fixed <- private$fixed_parameters$initial()
+        t(rbind(
+          matrix(fixed, length(names(fixed)), length(self$populations()),
+            dimnames = list(names(fixed), self$populations())),
+          varied))
```
richfitz, line 195: This looks more complicated than expected

R/pmcmc_nested_parameters.R
```diff
@@ -0,0 +1,488 @@
...
+    ##'
+    ##' @param theta a parameter matrix with columns in the same order as
+    ##' `$names()` and rows in the same order as `$populations()`.
+    prior = function(theta) {
+      priors <- numeric(nrow(theta))
+      names(priors) <- self$populations()
+
+      if (!is.null(private$fixed_parameters)) {
+        lp_fix <- private$fixed_parameters$prior(
+          theta[1, private$param_names$fixed])
```
richfitz, line 209: can you add a comment here saying that the `1` is arbitrary?

R/pmcmc_nested_parameters.R
```diff
@@ -0,0 +1,488 @@
...
+          varied))
+      }
+    },
+
+    ##' @description Compute the prior(s) for a parameter vector/matrix
+    ##'
+    ##' @param theta a parameter matrix with columns in the same order as
+    ##' `$names()` and rows in the same order as `$populations()`.
+    prior = function(theta) {
+      priors <- numeric(nrow(theta))
```
richfitz, line 204: This pair of lines uses two sources of truth without checking that they agree (you take the length from nrow(thta) but the names from the number of populations)

R/pmcmc_nested_parameters.R
```diff
@@ -0,0 +1,488 @@
...
+    ##' parameters were defined in (see `$names()` for that order.
+    model = function(theta) {
+      apply(theta, 1, private$transform)
+    },
+
+    ##' @description Set some parameters to fixed values. Use this to
+    ##' reduce the dimensionality of your system.
+    ##'
+    ##' @param fixed a named vector of parameters to fix
+    # TODO
```
richfitz, line 289: If you want to not implement this yet, please just add an assertion in the body and delete the code

R/pmcmc_nested_parameters.R
```diff
@@ -0,0 +1,488 @@
...
+    function(x) identical(names(x), populations))
+  if (!all(ok)) {
+    stop("Populations and ordering of varied parameters should be identical.")
+  }
+
+  list(fixed = fixed_parameters, varied = varied_parameters, pop = populations,
+    names = list(fixed = fix_nms, varied = var_nms))
+}
+
+clean_proposals <- function(proposal_fixed, proposal_varied, populations,
```
richfitz, line 392: I think the code below could be clarified if you lift things like `length(params$varied)` into constants at the top of the functions as you use them several times in complicated if statements

richfitz (approved):
From the point of view of the `pmcmc_nested_parameters` object, the interface is great. We can tidy up details later.

Things that you should look at here:

* pmcmc_nested_parameters; I wonder if we might name pmcmc_parameters_nested
* there are small formatting issues here; you might reflow this in your editor and see if they get dealt with. I know it's picky, but these things build up over time and then we get noisy diffs that are hard to deal with
* the testing might be tidied up a bit; I recognise the testing style, but the issue is that some of these blocks test all sorts of things which will make the tests hard to maintain. I am happy to tidy some of that up later.
* the docs need writing :)

richfitz (changes requested):
I swear I hit the wrong button 30% of the time

R/pmcmc_nested_parameters.R
```diff
@@ -0,0 +1,488 @@
...
+
+  type <- c("pmcmc_parameter", "pmcmc_varied_parameter")
+  ok <- vlapply(parameters, inherits, type)
+  if (!all(ok)) {
+    stop(sprintf("Expected all elements of '...' to be '%s' objects", type))
+  }
+
+  varied <- vlapply(parameters, inherits, "pmcmc_varied_parameter")
+  if (!any(varied)) {
+    stop("No varied parameters, use `pmcmc_parameters` instead.")
```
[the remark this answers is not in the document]
RaphaelS1: I can change it now, it would just be mirroring the varied only case

R/pmcmc_nested_parameters.R
```diff
@@ -0,0 +1,488 @@
...
+  # check fixed and varied names separately
+  if (!is.null(fixed_parameters)) {
+    fix_nms <- vcapply(fixed_parameters, "[[", "name", USE.NAMES = FALSE)
+  } else {
+    fix_nms <- c()
+  }
+
+  # only need to get the name of the first population as internally asserted
+  var_nms <- vcapply(varied_parameters, function(x) x[[1]]$name[1],
+    USE.NAMES = FALSE)
```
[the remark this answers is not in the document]
RaphaelS1: My editor is not happy. I can manually change these though

R/pmcmc_nested_parameters.R
RaphaelS1: I can tidy up

R/pmcmc_nested_parameters.R
RaphaelS1: That was my logic too, i.e. as dust requires it then this should always be the case

R/pmcmc_nested_parameters.R
RaphaelS1: Really? It is stored as NULL if not present

richfitz: I'd store it as `list()` if not present (i.e., none are present)

R/pmcmc_varied_parameter.R
```diff
@@ -0,0 +1,77 @@
...
+pmcmc_varied_parameter <- function(name, populations, initial, min = -Inf,
+                                   max = Inf, discrete = FALSE, prior = NULL) {
+
+  assert_character(populations)
+  len <- length(populations)
+
+  initial <- recycle(initial, len)
+  min <- recycle(min, len)
+  max <- recycle(max, len)
+  discrete <- recycle(discrete, len)
```
richfitz, line 59: This should not be recyclable; must be scalar

R/utils.R
```diff
@@ -12,7 +12,6 @@ vnapply <- function(x, fun, ...) {
   vapply(x, fun, numeric(1), ...)
 }
 
-
```
richfitz, line 15: Following pep8 style I usually leave 2 lines between functions; it's never that consistent...

R/pmcmc_parameters_nested.R
```diff
@@ -0,0 +1,594 @@
...
+
+    ##' @description Return the initial parameter values as a named matrix with
+    ##' rows corresponding to populations and columns are parameters.
+    initial = function() {
+
+      varied_params <- private$param_names$varied
+      fixed_params <- private$param_names$fixed
+      pops <- self$populations()
+
+      if (length(varied_params)) {
```
richfitz, line 202:
```suggestion
      if (length(varied_params) > 0) {
```

(and elsewhere)

R/pmcmc_parameters_nested.R
```diff
@@ -0,0 +1,594 @@
...
+      if (is.null(rownames(fixed)) || is.null(colnames(fixed))) {
+        stop("`fixed` should have rownames (populations) and colnames
+             (parameters).")
+      }
+
+      param_names <- private$param_names$original
+      pops <- self$populations()
+
+      if (any(is.na(match(rownames(fixed), pops)))) {
+        stop("Rownames of `fixed` should be identical to `$populations`.")
```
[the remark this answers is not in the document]
richfitz: also we use single quotes, not back quotes in these messages, and usually do not include a period

```suggestion
        stop("Rownames of 'fixed' should be identical to '$populations'")
```

R/pmcmc_parameters_nested.R
```diff
@@ -0,0 +1,594 @@
...
+make_proposal_fixed <- function(kernel) {
+  rmv <- rmvnorm_generator(kernel)
+  nms <- rownames(kernel)
+  function(mean, scale = 1.0) {
+    mean[, nms] <- rep(rmv(mean[1, nms], scale = scale), each = nrow(mean))
+    mean
+  }
+}
+
+get_inner_params <- function(param, population, private) {
```
richfitz, line 578: Hmm, I wonder if we might tidy this up later - some way of propagating the information back via the summary method? Doesn't need to be done now

richfitz (changes requested):
Some of the corner-case catching (involving `t()`) could be tightened up, but we can also leave that for later as it's a detail. The `discrete` replication and single remaining coverage hole are probably the most important

R/pmcmc_parameters_nested.R
```diff
@@ -0,0 +1,594 @@
...
+    propose = function(theta, scale = 1, type = c("fixed", "varied", "both")) {
+      type <- match.arg(type)
+      pfix <- function(theta, scale) {
+        if (length(private$proposal_fixed)) {
+          private$proposal_fixed(theta, scale)
+        }
+      }
+      pvar <- function(theta, scale) {
+        if (length(private$proposal_varied)) {
+          var <- vapply(seq(nrow(theta)), function(i)
```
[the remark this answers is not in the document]
RaphaelS1: sorry not sure which pattern you mean

R/pmcmc_parameters_nested.R
```diff
@@ -0,0 +1,594 @@
...
+      if (length(fixed_params)) {
+        fixed <- private$fixed_parameters$initial()
+        fixed <- matrix(fixed, length(fixed_params), length(pops),
+                        dimnames = list(fixed_params, pops))
+      } else {
+        fixed <- NULL
+      }
+
+      init <- t(rbind(fixed, varied))
+      init[, match(self$names(), colnames(init)), drop = FALSE]
```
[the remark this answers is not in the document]
RaphaelS1: Do you mean saving the matched indices?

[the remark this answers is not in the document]
richfitz: the entire initial conditions. It's nbd and we can explore that later if we want to

R/pmcmc_parameters_nested.R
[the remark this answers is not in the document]
richfitz: we're getting a series of vectors built by vapply into a matrix and special-casing the single-parameter version. We can tidy that up later though

richfitz (approved):

PR #109, opened 2021-02-17
Bugfix for discrete and bounded nested parameters

RaphaelS1:


R/pmcmc_parameters_nested.R
```diff
@@ -43,45 +43,43 @@ pmcmc_parameters_nested <- R6::R6Class(
...
-    ##' [pmcmc_varied_parameter] objects are included in `parameters`.
-    ##' Otherwise population names are taken from those objects.
-    ##'
-    ##' @param transform An optional transformation function to apply
-    ##' to your parameter vector immediately before passing it to the
-    ##' model function. If not given, then [as.list] is
-    ##' used, as dust models require this. However, if you need to
-    ##' generate derived parameters from those being actively sampled
-    ##' you can do arbitrary transformations here.
+##' @description Create the pmcmc_parameters object
```
richfitz, line 55: The previous indenting here was correct - could you reflow?

R/pmcmc_parameters_nested.R
```diff
@@ -247,73 +237,58 @@ pmcmc_parameters_nested <- R6::R6Class(
...
+##' proposal distribution. This may be useful in sampling starting
+##' points. The parameter is equivalent to a multiplicative factor
+##' applied to the variance covariance matrix.
+##'
+##' @param type specifies which type of parameters should be proposed,
+##' either fixed parameters only ("fixed"), varied only ("varied"), or
+##' both ("both") types. For 'fixed' and 'varied', parameters of the
+##' other type are left unchanged.
+    propose = function(theta, scale = 1, type = c("both", "varied", "fixed")) {
       type <- match.arg(type)
```
richfitz, line 266: What do you think about making this required, rather than with a default?

Normally I write a stricter version of match.arg that requires exact matches

R/pmcmc_parameters_nested.R
```diff
@@ -247,73 +237,58 @@ pmcmc_parameters_nested <- R6::R6Class(
...
-      pfix <- function(theta, scale) {
-        if (length(private$proposal_fixed) > 0) {
-          private$proposal_fixed(theta, scale)
-        }
+      fix_params <- private$param_names$fixed
+      var_params <- private$param_names$varied
+
+      nok <- (type == "fixed" && length(fix_params) == 0) ||
```
richfitz, line 270: This is fine for now, but we might check with Marc/Ed about whether the correct behaviour is for this to act like an identity function on theta

R/pmcmc_parameters_nested.R
```diff
@@ -247,73 +237,58 @@ pmcmc_parameters_nested <- R6::R6Class(
...
-          # catch one varied edge case
-          if (!inherits(var, "matrix")) {
-            var <- as.matrix(var)
-            rownames(var) <- private$param_names$varied
-          }
 
-          colnames(var) <- self$populations()
-          t(var)
-        }
+      # arbitrarily select first population as both same
```
richfitz, line 276: While we get this up and running, we might pop in a couple of assertions that this is true, to guard from bookkeeping failure. Can do this later, (or not at all)

richfitz (changes requested):
This looks heaps simpler. Some suggestions included

R/pmcmc_parameters_nested.R
RaphaelS1: Sure I'll add these

R/pmcmc_parameters_nested.R
RaphaelS1: I had also considered this. However I was put off by the case of trying to update parameters over 1000 steps that are never updated, e.g. only fixed exist and request fixed update only.

richfitz (approved):

PR #111, opened 2021-02-19
Extend particle filter for populations

RaphaelS1:
* Adds `particle_filter_data_multi` and extends `particle_filter_data` to handle populations


R/utils.R
```diff
@@ -116,3 +133,88 @@ recycle <- function(x, n, name = deparse(substitute(x))) {
...
+    d[3L]
+  } else {
+    1L
+  }
+}
+
+
+## Copied from colnames
+layernames <- function(x, do.NULL = TRUE, prefix = "layer") { # nolint
+    dn <- dimnames(x)
```
richfitz, line 161: can you re-indent please

R/utils.R
```diff
@@ -116,3 +133,88 @@ recycle <- function(x, n, name = deparse(substitute(x))) {
...
+    if (!is.null(dn[[3L]]))
```
richfitz, line 162: Please always use braces, even for one-line if / else clauses

R/utils.R
```diff
@@ -116,3 +133,88 @@ recycle <- function(x, n, name = deparse(substitute(x))) {
...
+  if (length(d) > 2L) {
+    d[3L]
+  } else {
+    1L
+  }
+}
+
+
+## Copied from colnames
+layernames <- function(x, do.NULL = TRUE, prefix = "layer") { # nolint
```
richfitz, line 160: would we ever use the do.NULL argument? I don't recognise it from base (as in, I see it's there but have never seen it before)

R/utils.R
```diff
@@ -116,3 +133,88 @@ recycle <- function(x, n, name = deparse(substitute(x))) {
...
+      if (do.NULL) {
+        NULL
+      } else {
+        paste0(prefix, seq_len(NLAYER(x)))
+      }
+    }
+}
+
+
+## Copied from colnames<-
```
richfitz, line 174: rather than a copy (with copyright issues) can you adapt this to the subset that we need perhaps?

R/utils.R
```diff
@@ -116,3 +133,88 @@ recycle <- function(x, n, name = deparse(substitute(x))) {
...
+  x
+}
+
+
+set_layernames <- function(m, nms) {
+  layernames(m) <- nms
+  m
+}
+
+lbind <- function(arrays) {
```
richfitz, line 207: this is array_bind I think, though that is likely more efficient as it avoids the do.call; binding on the last array dimension is the easiest

R/utils_assert.R
```diff
@@ -88,3 +88,16 @@ assert_scalar_character <- function(x, name = deparse(substitute(x))) {
   assert_character(x, name)
   invisible(x)
 }
+
+assert_list <- function(x, class = NULL, name = deparse(substitute(x))) {
```
richfitz, line 92: assert_list_of perhaps, with no default handling of `class`

richfitz (changes requested):
This looks great - mostly pretty small comments tbh given the scale of the change

R/particle_filter_data.R
```diff
@@ -31,11 +31,22 @@
 ##'   must be at most equal to the first value of the `time`
 ##'   column, minus 1 (i.e., `data[[time]] - 1`).
 ##'
-##' @return A data.frame with new columns `step_start` and
-##'   `step_end` (required by [`particle_filter`]),
+##' @param population Optionally, the name of a column within `data` that
+##'   represents different populations. This column must be coercable to
+##'   factor.
```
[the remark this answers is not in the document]
RaphaelS1: Agreed which is why here I make it forced to already be factor. The docs are wrong but code is forced factor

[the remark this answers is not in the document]
richfitz: ah great - was going off the docs and never cross checked the code

R/particle_filter.R
```diff
@@ -288,28 +318,71 @@ particle_filter <- R6::R6Class(
...
-      }
-
       ny <- nrow(history_value)
-      np <- length(index_particle)
-      nt <- ncol(history_order)
 
-      idx <- matrix(NA_integer_, np, nt)
-      for (i in rev(seq_len(ncol(idx)))) {
-        index_particle <- idx[, i] <- history_order[index_particle, i]
+      if (length(dim(history_value)) == 4) {
```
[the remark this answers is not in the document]
RaphaelS1: Not sure if we can simplify that much. In the other case it doesn't matter how long `index_particle` is (and we don't check here) but it does matter if a matrix is provided with columns not corresponding to number of populations; here we also offer automated coercion to matrix with same indices for all populations, but there's no need to do this (i.e. user could supply different indices for each pop). I can factor into free function but not sure what can be trimmed

R/particle_filter.R
```diff
@@ -171,17 +172,33 @@ particle_filter <- R6::R6Class(
...
+      if (inherits(data, "particle_filter_data_nested")) {
+        if (is.null(compare)) {
+          private$data_split <- dust::dust_data(private$data, "step_end",
+                                                multi = "population")
+        } else {
+          private$data_split <- groupeddf_to_list_of_lists(data, "population")
+        }
+        private$steps <- unname(
+          as.matrix(
+            split(data, data$population)[[1]][c("step_start", "step_end")]
```
richfitz, line 185: should we subset rather than split here?

R/particle_filter.R
```diff
@@ -473,3 +500,73 @@ check_save_restart <- function(save_restart, data) {
...
+
+  cidx <- cbind(seq_len(ny),
+                rep(idx, each = ny),
+                rep(seq_len(nt), each = ny * np))
+  ret <- array(history_value[cidx], c(ny, np, nt))
+  rownames(ret) <- names(history_index)
+  ret
+}
+
+history_nested <- function(history_value, history_order, history_index,
```
richfitz, line 531: Thanks for breaking this out. I implemented this a little while ago in dust (https://github.com/mrc-ide/dust/blob/ed782f0e9ace0422cc88fd2bd8dc7dc843183947/inst/include/dust/dust.hpp#L127-L165) and given how nasty it is we might centralise in a future PR so that this logic only exists in one place. This will be much easier to do with your change here

R/particle_filter_data.R
```diff
@@ -69,6 +97,95 @@ particle_filter_data <- function(data, time, rate, initial_time = NULL) {
...
+  t <- clean_pf_times(split_data[[1]][[time]], initial_time, data)
+
+  data <- do.call(rbind, split_data)
+
+  ret <- data.frame(time_start = t$time_start,
+                    time_end = t$time_end,
+                    step_start = t$time_start * rate,
+                    step_end = t$time_end * rate,
+                    population = rep(groups, each = length(t$time_start)),
+                    data[vars])
```
richfitz, line 162: stringsAsFactors = FALSE here to prevent possible coersion of strings in `data[vars]` in R < 4

richfitz (changes requested):
ok tiny things here, then this is good to go and we can start back on #108

richfitz (approved):

richfitz:
(merging early as the previous failure unrelated to this PR and will most likely pass)

PR #116, opened 2021-02-24
Add vignette for nested models

RaphaelS1:


scripts/generate_vignette_data.R
```diff
@@ -8,7 +8,7 @@ set.seed(0)
 gen_sir <- dust::dust_example("sir")
 dt <- 0.25
 inv_dt <- 1 / dt
-sir <- gen_sir$new(data = list(dt = dt, S_ini = 1000, I_ini = 10,
+sir <- gen_sir$new(pars = list(dt = dt, S_ini = 1000, I_ini = 10,
```
richfitz, line 11: thanks!

richfitz (approved):
This is great. Docs can always be improved and extended so when we find things that want adding to it we can do that too

PR #124, opened 2021-03-02
fix particle_filter_state_nested initial

RaphaelS1:


R/predict.R
```diff
@@ -51,25 +51,67 @@ pmcmc_predict <- function(object, steps, prepend_trajectories = FALSE,
...
+  
+  if (is_3d_array(state)) {
+    res <- pmcmc_predict_nested(object, state, index, model, n_threads, 
+                                steps, prepend_trajectories, seed)
+  } else {
+    pars <- apply(object$pars, 1, object$predict$transform) 
+    
+    ## NOTE: n_particles here is 1 because every particle gets a
+    ## different state.
+    mod <- model$new(pars, steps[[1]], 1L, n_threads = n_threads,
```
richfitz, line 66: with new dust this can be relaxed to be NULL, I think that's what you want

richfitz (commented):
one comment that I think is important, will review the test later

R/particle_filter_state_nested.R
```diff
@@ -83,31 +84,55 @@ particle_filter_state_nested <- R6::R6Class(
       }
 
       if (!is.null(initial)) {
-        initial_data <- initial(model$info(), n_particles, pars)
-        if (is.list(initial_data)) {
-          steps <- particle_steps(steps, initial_data$step)
-          model$set_state(initial_data$state, initial_data$step)
+        initial_data <- Map(initial,
```
richfitz, line 87: Can you pull all this logic (probably everything from and including the if statement) into a free function to make it easier to contain please?

R/particle_filter_state_nested.R
```diff
@@ -83,31 +84,55 @@ particle_filter_state_nested <- R6::R6Class(
...
+          info = model$info(),
+          pars = pars,
+          MoreArgs = list(n_particles = n_particles)
+        )
+        if (!is.null(unlist(initial_data))) {
+          if (any(c(c("step", "state") %in% names(initial_data[[1]])))) {
```
richfitz, line 93: Generally we probably want to not just check the first element but all. So perhaps something like:

```
initial_step <- lapply(initial_data, "[[", "step")
initial_state <- lapply(initial_data, "[[", "state")
```

then work with these lists

R/particle_filter_state_nested.R
```diff
@@ -83,31 +84,55 @@ particle_filter_state_nested <- R6::R6Class(
...
+            init_step <- unlist(lapply(initial_data, "[[", "step"))
+            if (length(init_step) != n_particles * length(pars)) {
+              if (length(init_step) == length(pars)) {
+                init_step <- rep(init_step, each = n_particles)
+              } else {
+                stop(sprintf("Expected 'step' to be length %d or %d",
```
richfitz, line 99: This error is probably not user-actionable; if the initial function gives unexpected outputs here it'll be hard to debug. So probably it needs adjusting so that it says something like: 

```
"initial() produced unexpected step length %d for parameter set %d (expected %d or %d)"
```

but then it might be easiest to manage that like:

```
initial_step_len <- lengths(init_step) # lengths is one of my fav underknown functions
if (length(unique(initial_step_len)) != 1) {
  stop("initial() produced unequal step lengths ...") # also provide some information about what was found
}
initial_step_len <- initial_step_len[[1]] # this is safe now because we know they're all exchangable
if (initial_step_len == 1) {
  # build your matrix by repetition
} else if (initial_step_len == n_particles) {
  # build your matrix by assembling it
} else {
  stop(sprintf("initial() produced unexpected step length %d (expected 1 or %d)",
               initial_step_len, n_particles))
}
```

R/particle_filter_state_nested.R
```diff
@@ -83,31 +84,55 @@ particle_filter_state_nested <- R6::R6Class(
...
         } else {
-          model$set_state(initial_data)
+          model$set_state(NULL)
         }
       }
 
-      index_data <- if (is.null(index)) NULL else index(model$info())
-      if (!is.null(compare) && !is.null(index_data$run)) {
-        model$set_index(index_data$run)
+      index_data <- if (is.null(index)) NULL else lapply(model$info(), index)
```
richfitz, line 117: We probably should check that all index entries are the same

R/particle_filter_state_nested.R
```diff
@@ -208,18 +233,19 @@ particle_filter_state_nested <- R6::R6Class(
         if (is.null(compare)) {
           log_weights <- model$compare_data()
         } else {
-          ok <- !is.null(compare(array(state[, , 1],
-                                       c(nrow(state), ncol(state))),
-                                 data_split[[t]][[1]], pars))
+          ok <- !is.null(compare(
```
richfitz, line 236: Why are we running compare() twice here?

R/pmcmc_state.R
```diff
@@ -337,11 +337,12 @@ pmcmc_state <- R6::R6Class(
         ## [state x particle x population x step]
         trajectories_state <-
           list_to_array(private$history_trajectories$get())
-        trajectories_state <- aperm(trajectories_state, c(1, 3, 2, 4))
+        trajectories_state <- aperm(trajectories_state, c(1, 4, 2, 3))
         rownames(trajectories_state) <- names(predict$index)
 
         data <- private$filter$inputs()$data
-        step <- c(data$step_start[[1]], data$step_end)
+        step_end <- data$step_end[seq_len(table(data$population)[1])]
```
richfitz, line 344: What is the role of table here?

richfitz (changes requested):
A mix of suggestions and coding hints here

R/particle_filter_state_nested.R
RaphaelS1: I don't know if the current setting allows one element to have 'step' or 'state' and not the other. The `initial` function (at least not in tests or carehomes) doesn't differentiate between what (step/state/etc) to return for populations, just the values they take.

R/pmcmc_state.R
```diff
@@ -320,15 +320,15 @@ pmcmc_state <- R6::R6Class(
...
         state <- list_to_array(private$history_state$get())
         colnames(state) <- rownames(private$curr_pars)
       }
 
       if (length(private$control$save_restart) > 0) {
-        ## [state x pop x time]
+        ## [state x sample x pop x time]
         restart_state  <-
-          list_to_array(private$history_restart$get())
+          aperm(list_to_array(private$history_restart$get()), c(1, 4, 2, 3))
```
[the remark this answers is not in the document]
RaphaelS1: This is where it was breaking before. This should align it correctly with un-nested case, but I think we'll probably need to go through every index properly at some point to confirm

R/pmcmc.R
```diff
@@ -101,7 +101,7 @@ pmcmc_multiple_series <- function(pars, initial, filter, control) {
 
   for (i in seq_along(samples)) {
     if (control$progress) {
-      message(sprintf("Running chain %d / %d", i, control$n_chains))
+      message(sprintf("\nRunning chain %d / %d", i, control$n_chains))
```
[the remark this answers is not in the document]
RaphaelS1: I actually forgot I had done this. On my Windows this was just printing each on the same line and looked a bit messy, so I added newline

R/particle_filter_state_nested.R
RaphaelS1: As with the `initial` comment above. Not sure if we currently allow a case where index entries could be different between populations

R/particle_filter_state_nested.R
RaphaelS1: No idea....

R/pmcmc_state.R
RaphaelS1: Gets the number of rows in each population

R/particle_filter_state_nested.R
RaphaelS1: Oh! Because of the `vapply` https://github.com/mrc-ide/mcstate/blob/e4798e9907f41f46bda0cf24459b316b385232cc/R/particle_filter_state_nested.R#L243-L248 . I guess in testing I found it could return NULL for some `t` but not all, and running the `vapply` threw errors when this was the case

R/particle_filter_state_nested.R
```diff
@@ -83,31 +84,55 @@ particle_filter_state_nested <- R6::R6Class(
...
-          steps <- particle_steps(steps, initial_data$step)
-          model$set_state(initial_data$state, initial_data$step)
+        initial_data <- Map(initial,
+          info = model$info(),
+          pars = pars,
+          MoreArgs = list(n_particles = n_particles)
+        )
+        if (!is.null(unlist(initial_data))) {
+          if (any(c(c("step", "state") %in% names(initial_data[[1]])))) {
+            init_step <- unlist(lapply(initial_data, "[[", "step"))
```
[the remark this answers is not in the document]
RaphaelS1: Do you have a suggestion for how to avoid the case when elements may be NULL though? `vapply` won't work, is `simplify2array(lapply...` preferred?

R/pmcmc.R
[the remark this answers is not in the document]
richfitz: We should fix this in the callback at the end of the progress bar run instead I think

R/pmcmc_state.R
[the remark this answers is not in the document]
richfitz: I get aperm wrong almost all the time

R/particle_filter_state_nested.R
richfitz: yes, either could be set but if any parameter sets one all must set it

R/particle_filter_state_nested.R
richfitz: We should not allow it, but enforce that they're identical unless that is too costly

R/particle_filter_state_nested.R
richfitz: in which case, run the main loop in an lapply, then a second vapply to collect together

R/pmcmc_state.R
richfitz: is this not constant?

`tabulate` is likely faster and less magic

richfitz (approved):

PR #125, opened 2021-03-08
Nested step ratio

RaphaelS1:
Adds `nested_step_ratio` parameter to `pmcmc_control` for controlling ratio of fixed:varied steps

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,6 @@
 # mcstate 0.5.9
 
-* Add `nested_step_ratio` parameter to `pmcmc_control` for controlling the ratio of fixed:varied
-steps for nested pMCMC
+* Add `nested_step_ratio` parameter to `pmcmc_control` for controlling the ratio of fixed:varied steps for nested pMCMC
 
 # mcstate 0.5.5
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# mcstate 0.5.9
+
+* Add parameter to `pmcmc_control` for controlling the ratio of fixed:varied
+steps for nested pMCMC
+
 # mcstate 0.5.5
 
 * New array helper `mcstate::array_flatten` for unshaping an array
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # mcstate 0.5.9
 
-* Add parameter to `pmcmc_control` for controlling the ratio of fixed:varied
+* Add `nested_step_ratio` parameter to `pmcmc_control` for controlling the ratio of fixed:varied
 steps for nested pMCMC
 
 # mcstate 0.5.5
--- a/R/pmcmc_control.R
+++ b/R/pmcmc_control.R
@@ -153,34 +153,34 @@ pmcmc_control <- function(n_steps, n_chains = 1L, n_threads_total = NULL,
         n_threads_total, n_workers))
     }
   }
-  
+
   if (!identical(unname(rerun_every), Inf)) {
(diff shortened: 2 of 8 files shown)


NEWS.md
```diff
@@ -1,3 +1,8 @@
+# mcstate 0.5.9
+
+* Add `nested_step_ratio` parameter to `pmcmc_control` for controlling the ratio of fixed:varied
```
richfitz, line 3: .md files are one exception to 80 char rule - we get most consistent formatting by soft-wrapping

R/pmcmc_control.R
```diff
@@ -167,6 +174,13 @@ pmcmc_control <- function(n_steps, n_chains = 1L, n_threads_total = NULL,
     assert_strictly_increasing(save_restart)
   }
 
+  nok <- !test_integer(nested_step_ratio) &&
+    !test_integer(1 / nested_step_ratio)
+  if (nok) {
```
richfitz, line 179:
```suggestion
  ok <- test_integer(nested_step_ratio) || test_integer(1 / nested_step_ratio)
  if (!ok) {
```

(fewer negations)

R/pmcmc_state.R
```diff
@@ -220,11 +229,20 @@ pmcmc_state <- R6::R6Class(
           private$update_history_nested()
         }
 
-        if (i %% 2) {
-          private$run_fixed()
+        if (!is.null(fix_vary_step)) {
```
richfitz, line 232: This (combined with the above) is pretty gnarly.

I wonder if you could pull it into a closure so that the interface would look like:

```r
run <- alternate(private$run_fixed, private$run_varied, private$control$nested_step_ratio)
...
run(j)
```

by doing something like (untested!):

```r
alternate <- function(f, g, ratio) {
  if (ratio < 1) {
    return(alternate, g, f, 1 / ratio)
  }
  assert_integer(ratio)
  function(i) {
    if (i %% (ratio + 1) == 0) {
      g()
    } else {
      f()
    }
  }
}
```

richfitz (changes requested):
this scale of PR is so much easier to review :pray:

R/pmcmc_state.R
```diff
@@ -220,11 +225,8 @@ pmcmc_state <- R6::R6Class(
           private$update_history_nested()
         }
 
-        if (i %% 2) {
-          private$run_fixed()
-        } else {
-          private$run_varied()
-        }
+        run_alternate_step(j)
+        j <- j + 1
```
richfitz, line 229: Do we need `j` now? `i` seems like it'd be the natural choice to pass in here and will prevent weirdness around restarting changing the effective ratio

R/pmcmc_control.R
```diff
@@ -99,6 +99,12 @@
 ##' @param progress Logical, indicating if a progress bar should be
 ##'   displayed, using [`progress::progress_bar`].
 ##'
+##' @param nested_step_ratio Either integer or 1/integer, which specifies the
+##'   ratio of fixed:varied steps in a nested pMCMC. For example `3` would run
+##'   3 steps proposing fixed parameters only and then 1 step proposing varied
+##'   parameters only; whereas `1/3` would run 3 varied steps
+##'   for every 1 fixed step.
```
richfitz, line 106: Add something like:

> The default value of 1 runs an equal number of iterations updating the fixed and varied parameters.

You might also want to provide guidance as to how to select a sensible value here based on the number of fixed and varied parameters

richfitz (changes requested):

richfitz (approved):

PR #127, opened 2021-03-12
Use compiled particle filter if given a compiled compare function

### LINKED ISSUE #118: Run compiled version of particle filter if available

richfitz:
This PR starts us towards using dust's particle filter engine rather than the pure R one. This should be slightly more efficient in copying and also runs the compare function in parallel. It will be interesting to see what the gain is on sircovid but I think there was scope for about a 10% improvement there (so ok, but not great). This is also needed to move us towards longer term work where the particle filter will be available on the GPU.

There's some tension here with supporting the partial run (itself for SMC^2). To do this in dust requires that we implement https://github.com/mrc-ide/dust/issues/177 - once that's done things might actually get easier

Fixes #118

### DIFF
--- a/R/particle_filter.R
+++ b/R/particle_filter.R
@@ -513,26 +513,32 @@ check_save_restart <- function(save_restart, data) {
 
 history_single <- function(history_value, history_order, history_index,
                            index_particle) {
-
   ny <- nrow(history_value)
 
-  if (is.null(index_particle)) {
-    index_particle <- seq_len(ncol(history_value))
+  if (is.null(history_order)) {
+    if (is.null(index_particle)) {
+      ret <- history_value
+    } else {
+      ret <- history_value[, index_particle, , drop = FALSE]
+    }
+  } else {
+    if (is.null(index_particle)) {
+      index_particle <- seq_len(ncol(history_value))
+    }
+
+    np <- length(index_particle)
+    nt <- ncol(history_order)
+
+    idx <- matrix(NA_integer_, np, nt)
+    for (i in rev(seq_len(ncol(idx)))) {
+      index_particle <- idx[, i] <- history_order[index_particle, i]
+    }
+
+    cidx <- cbind(seq_len(ny),
+                  rep(idx, each = ny),
+                  rep(seq_len(nt), each = ny * np))
+    ret <- array(history_value[cidx], c(ny, np, nt))
   }
-
-  np <- length(index_particle)
-
-  nt <- ncol(history_order)
-
-  idx <- matrix(NA_integer_, np, nt)
-  for (i in rev(seq_len(ncol(idx)))) {
-    index_particle <- idx[, i] <- history_order[index_particle, i]
-  }
-
-  cidx <- cbind(seq_len(ny),
-                rep(idx, each = ny),
-                rep(seq_len(nt), each = ny * np))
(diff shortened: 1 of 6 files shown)


richfitz:
For sircovid:

```r
  start_date <- sircovid_date("2020-02-02")
  pars <- carehomes_parameters(start_date, "england")
  data <- carehomes_data(read_csv(sircovid_file("extdata/example.csv")),
                         start_date, pars$dt)

  pf <- carehomes_particle_filter(data, 100, 10)
  pf2 <- carehomes_particle_filter(data, 100, 10, compiled_compare = TRUE)

  bench::mark(
    pf$run(pars),
    pf2$run(pars),
    pf$run(pars, TRUE),
    pf2$run(pars, TRUE),
    check = FALSE,
    min_iterations = 10,
    memory = FALSE)
```

gives:

```
# A tibble: 4 x 13
  expression            min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>          <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 pf$run(pars)        132ms  135ms      6.98        NA    0.776     9     1
2 pf2$run(pars)       127ms  129ms      7.74        NA    0        10     0
3 pf$run(pars, TRUE)  135ms  140ms      7.01        NA    3.00      7     3
4 pf2$run(pars, TRUE) 129ms  130ms      7.61        NA    0.845     9     1
```

so perhaps a 7% speedup, a little more pronounced when collecting history. It's possible the advantage will become a bit more pronounced with longer model runs but this is about inline with my memories.

R/particle_filter.R
```diff
@@ -513,68 +513,93 @@ check_save_restart <- function(save_restart, data) {
...
       }
+      d <- dim(history_value)
+      d[[2L]] <- nrow(index_particle)
+      ret <- array(NA_real_, d)
+      for (i in seq_len(npop)) {
+        ret[, , i, ] <- history_value[, index_particle[, i], i, ]
+      }
+    }
+  } else {
+    npop <- ncol(history_order)
```
RaphaelS1, line 568: Confusing that here it's ncol but nlayer above.. is this intentional?

R/particle_filter_state.R
```diff
@@ -191,6 +183,40 @@ particle_filter_state <- R6::R6Class(
       log_likelihood <- self$log_likelihood
       n_particles <- private$n_particles
 
+      ## I think that this can be factored into a little utility
+      ## because it will be very shared with the multiple parameter
+      ## case too.
```
RaphaelS1, line 188: Agreed it looks very similar

R/particle_filter.R
```diff
@@ -513,68 +513,93 @@ check_save_restart <- function(save_restart, data) {
...
-  nt <- nlayer(history_order)
 
-
-  if (is.null(index_particle)) {
-    index_particle <- matrix(seq_len(ncol(history_value)),
-                             ncol(history_value), npop)
-  } else {
-    if (is.matrix(index_particle)) {
+  if (is.null(history_order)) {
+    npop <- nlayer(history_value)
```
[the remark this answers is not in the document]
richfitz: The history comes out of the compiled filter pre-sorted

RaphaelS1 (approved):

RaphaelS1 (approved):

PR #128, opened 2021-04-07
Add support for rerunning randomly

richfitz:
As requested by Neil.


R/pmcmc_control.R
```diff
@@ -58,6 +58,11 @@
 ##'   point before doing the comparison.  This may help "unstick"
 ##'   chains, at the cost of some bias in the results.
 ##'
+##' @param rerun_random Logical, indicating if `rerun_every`
```
RaphaelS1, line 61: I found this very hard to read as lots of the same word are repeated

RaphaelS1 (approved):

PR #130, opened 2021-04-19
New approach to running chains in separate processes

richfitz:
This PR exposes (and lightly refactors) some of the internals of the pmcmc so that it's straightforward to parallelise the mcmc onto different nodes in a way that makes the underlying rng deterministic


R/pmcmc.R
```diff
@@ -66,6 +65,59 @@ pmcmc <- function(pars, filter, initial = NULL, control = NULL) {
...
+##'
+##' @inheritParams pmcmc
+##' @export
+pmcmc_chains_prepare <- function(pars, filter, initial = NULL, control = NULL) {
+  assert_is(pars, c("pmcmc_parameters", "pmcmc_parameters_nested"))
+  assert_is(filter, "particle_filter")
+  assert_is(control, "pmcmc_control")
+
+  if (control$n_workers != 1) {
+    stop("'n_workers' must be 1")
```
RaphaelS1, line 82: Might be nice if this was explained in minimal detail in the document, seems counter-intuitive unless you know how it functions immediately

R/pmcmc.R
```diff
@@ -66,6 +65,59 @@ pmcmc <- function(pars, filter, initial = NULL, control = NULL) {
...
+  }
+  if (!control$use_parallel_seed) {
+    stop("'use_parallel_seed' must be TRUE")
```
RaphaelS1, line 85: Can we just auto-set this given it wouldn't break anything if it was unintentionally set to TRUE? (I don't think it would?)

R/pmcmc.R
```diff
@@ -66,6 +65,59 @@ pmcmc <- function(pars, filter, initial = NULL, control = NULL) {
 }
 
 
+##' Run a pMCMC, with sensible random number behaviour, but schedule
```
RaphaelS1, line 68: This function doesn't seem to be running anything but only preparing? Makes the docs a bit confusing

R/pmcmc.R
```diff
@@ -66,6 +65,59 @@ pmcmc <- function(pars, filter, initial = NULL, control = NULL) {
...
+}
+
+
+##' @param chain_id The chain index to run (1, 2, ..., `control$n_chains`)
+##'
+##' @param inputs A `pmcmc_inputs` object created by `pmcmc_chains_prepare`
+##'
+##' @export
+##' @rdname pmcmc_chains_prepare
+pmcmc_chains_run <- function(chain_id, inputs) {
```
RaphaelS1, line 109: Very minor and subjective point but I personally find it easier to read:
```R
pmcmc_chains_run(inputs, 1)
pmcmc_chains_run(inputs, 2)
```

than

```R
pmcmc_chains_run(1, inputs)
pmcmc_chains_run(2, inputs)
```

R/pmcmc.R
richfitz: Both functions are documented on the same page :)

R/pmcmc.R
richfitz: I'm not a fan of changing options behind the user's back because then you get odd dependencies between options. We might change the default at some point

R/pmcmc.R
richfitz: lapply disagrees with that, the current order is optimised to take advantage of that:

```
lapply(seq_len(nchains), pmcmc_chains_run, inputs)
```

the other way around you'd need to do

```
lapply(seq_len(nchains), pmcmc_chains_run, inputs = inputs) # or
lapply(seq_len(nchains), function(i) pmcmc_chains_run(inputs, i))
```

RaphaelS1 (approved):

PR #131, opened 2021-04-21
add one step nested update 

RaphaelS1:
Adds control `nested_alternate`. If FALSE then fixed & varied parameters proposed simultaneously and accepted/rejected collectively


richfitz:
Currently blocked on getting the sircovid nested run done

PR #133, opened 2021-05-14
Add IF2 algorithm

### LINKED ISSUE #123: Implement IF2 algorithm
See description here: https://kingaa.github.io/short-course/mif/mif.html

Started on branch `if2`: bb014d158ce09c9559099b211eb09fca1f764c7c
Requires dust to support a shuffle across different parameter sets

johnlees:
See:
- https://kingaa.github.io/pomp/vignettes/mif2.html
- https://kingaa.github.io/sbied/mif/if2_settings.html
- https://kingaa.github.io/sbied/mif/notes.pdf

TODO:
- [x] Make an IF2 control class
- [x] Add full set of tests for IF2 parameter class
- [x] Add full set of tests for IF2 class
- [x] Add full set of tests for IF2 control class
- [x] Docstrings (including examples)

Fixes #123


R/if2.R
```diff
@@ -0,0 +1,279 @@
...
+        stop("IF2 must be run first")
+      }
+      private$if_pars
+    },
+
+    ##' @description Plot results
+    ##'
+    ##' #param what Which parameter to plot. Any of `pars$names()` or `ll`
+    ##' for the log-likelihood
+    plot = function(what = "ll") {
```
richfitz, line 211: Generally, we've avoided plot methods, and I'd like to keep it that way because they're tarpits and hard to test

R/if2.R
```diff
@@ -0,0 +1,279 @@
+##' Run an IF2 inference. This is experimental and subject to change. Use at
```
richfitz: I think this warning is probably not needed

R/if2.R
```diff
@@ -0,0 +1,279 @@
...
+      log_likelihood <- rep(0, iterations)
+      if_pars <- array(NA_real_, c(n_pars, n_par_sets, iterations))
+      alpha_cool <- cooling_target ^ (1 / iterations)
+
+      p <- pmcmc_progress(iterations, private$control$progress)
+
+      for (m in seq_len(iterations)) {
+        p()
+        model$reset(pars = private$pars$model(pars_matrix), steps[[1L]])
+        ## TODO: refactor internal loop into function
```
richfitz, line 156: I think this is ok here

R/smc2_parameters.R
```diff
@@ -62,7 +62,7 @@ smc2_parameter <- function(name, sample, prior,
 }
 
 
-##' @title pmcmc_parameters
```
richfitz, line 65: thanks!

tests/testthat/test-if2.R
```diff
@@ -0,0 +1,143 @@
+context("IF2")
+
+test_that("Can run IF2", {
```
richfitz, line 3: could you add a test that shows it converges on approximately the right likelihood for the volatility case?

R/if2.R
```diff
@@ -0,0 +1,279 @@
...
+##' # Plot results
+##' filter$plot("ll")
+##' filter$plot("beta")
+##' filter$plot("gamma")
+##'
+##' # Get log-likelihood estimates from running a particle filter at
+##' # each final parameter estimate
+##' n_particles <- 100
+##' ll_samples <- filter$sample(n_particles)
+if2 <- R6::R6Class(
```
richfitz, line 43: This is quite a different interface to that presented by pmcmc and I'd like to harmonise if possible. There, we have the functional interface where calling the function `pmcmc` returns a plain list of class `mcstate_pmcmc` which has pars, likelihoods etc. On the way it uses a semi-exported object pmcmc_state but that's an implementation detail.

So here, the `if2` function should use the object much as you have it, but return all the requested information as a list. Unless there's some reason that's not possible, in which case please let me know.

R/if2.R
```diff
@@ -0,0 +1,279 @@
+##' Run an IF2 inference. This is experimental and subject to change. Use at
+##' your own risk.
+##'
+##' @title Run iterated filtering (IF2 algorithm)
+##'
+##' @description Create an IF2 object for running
+##'   and interacting with an IF2 inference.
+##'
+##' @export
```
richfitz, line 8: Looks like we need to redocument, these are not in the NAMESPACE yet

```
devtools::document()
```

R/if2.R
```diff
@@ -0,0 +1,279 @@
+##' Run an IF2 inference. This is experimental and subject to change. Use at
+##' your own risk.
+##'
```
richfitz, line 2: Please add a link/ref to the original paper here

richfitz (changes requested):
Looks good so far - as you say it's not a massive piece of implementation. Or perhaps we're both broken by the cuda stuff.

In addition to the comments, could you add a vignette, no matter how brief, showing using this on either the sir or volatility model?

richfitz (approved):

PR #141, opened 2021-08-06
Allow particle filter early exit

### LINKED ISSUE #138: Allow early termination of particle filter
In pmcmc run, we can compute the minimum log likelihood we would be willing to accept by computing the acceptance probability before we run. If this was passed to the filter as an optional arg, we could stop running it (and return -Inf) if we're going to drop under it. This will work if the likelihood only every decreases, which is the case in sircovid at least

richfitz:
If the log-likelihood calculation proceeds by adding only negative numbers, then when running a pmcmc we can tell when we would not be interested in the result as it will be certainly rejected. This PR implements support for this, with two components:

* an argument `min_log_likelihood` to the `mcstate::particle_filter`'s `$run()` method, which returns `-Inf` at the point where the likelihood becomes too small
* some logic changes to pmcmc to compute the minimum acceptable likelihood (this rearranges the usual MH acceptance step slightly)

~I've not run this on sircovid yet, but based on some experimentation I think we should see something like a 10% speed increased given the excessive variance~
Preliminary runs on sircovid suggest it's highly chain dependent, with an average of ~10% across 3 chains that I tried. Total configuration change is to set `filter_early_exit = TRUE` to the pmcmc_control call in spimalot.

~Builds on and so requires #140~
Fixes #138

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mcstate
 Title: Monte Carlo Methods for State Space Models
-Version: 0.6.3
+Version: 0.6.4
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("Marc", "Baguelin", role = "aut"),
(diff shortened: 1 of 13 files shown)


edknock:
Looks good, I appreciate that the documentation clearly states the option should only be used safely under certain distributions

edknock (approved):

PR #160, opened 2021-11-29
Basic multi-stage particle filter

### LINKED ISSUE #159: Multistage filter
Allow running a particle filter with several epochs. Between each epoch we may update parameters and state arbitrarily.

richfitz:
This PR implements a multi-stage particle filter, allowing arbitrary parameter and state transformations to occur at fixed time points along a time series.

The idea here is that we want to structurally change a model. Our use case is driven by the covid work:

1. run a basic model up to December
2. introduce vaccination (and eventually a second strain to model Alpha)
3. introduce a variant in ~May for Delta
4. swap strains (delta becomes strain 1, omicron comes in as strain 2)

Doing this will change the model "size" at both 1->2 and 2->3 as we increase the number of compartments. We don't want to run from the beginning of time with all these because they are huge, and even though most of the logic short circuits them it is much faster to not include them in the model. Doing the change requires increasing the size of the model by adding more layers to compartments (S, I, R becoming stratified by vaccination status and strain), people may need moving around a bit, and the parameters object needs replacing. Step 3->4 is a large change to the model state and also a change to the parameters, though the model state size will not change.

To allow this, I've set up a new type of parameters that captures fixed "epochs" where different parameters can be used, and between them where the state will be modified (see `multistage_epoch` and `multistage_parameters`).  Once in the filter things are fairly straightforward - conceptually we just run through each phase, and do some bookkeeping at the end.  

Assembling trajectories is a bit tricky, and for now I've set it up to use the names of the saved trajectories. We build a set of unique names and construct a final multdimensional array with NA values for state values that did not exist in some parts of the time.

The restart is a bit worse; for now at least I've asserted that we can only save restart state such that all restart times have the same state size. We could relax this but it would make this PR much harder to understand.

I believe that with this set up, it will work directly with `pmcmc` because we already use a user-provided transformation function to move from the real-vector of MCMC parameters to the list parameters used by the filter. So long as it returns a `multistage_parameters` object then it will use the multistage filter!

Things not done:

* deterministic filter does not support this yet; probably not that hard to add in, the bookkeeping is similar
* nested filter does not support this yet; better to refactor how that code is written as it's highly duplicated (#161)
* restart states must have the same size
* add a vignette showing this working properly (doing that really requires something like an SIR model with vaccination be written)

Fixes #159


edknock (approved):
Works pretty nicely!

PR #162, opened 2021-12-15
Start support for partial multistage filters

richfitz:
This is not really tested yet, and should not be merged until we carefully make sure it's doing the right thing!


richfitz:
Closed in favour of #164

PR #183, opened 2022-01-10
Allow filtering pmcmc on creation

richfitz:
This PR implements a new approach to filtering mcmc on creation. It is a *third* form of thinning unfortunately.

There is a small breaking change where we no longer output the initial step. I think that's ok though?


R/pmcmc_control.R
```diff
@@ -124,6 +152,13 @@
 ##'   calculation is a sum of discrete normalised probability
 ##'   distributions, but may not be for continuous distributions!
 ##'
+##' @param n_burnin Optionally, theumber of points to discard as
```
edknock, line 155:
```suggestion
##' @param n_burnin Optionally, the number of points to discard as
```

richfitz: updated in 1dfbe21 so that I could patch the generated docs too

edknock (approved):

PR #192, opened 2022-01-24
Allow use of compiled compare in more places

### LINKED ISSUE #177: Support for using the compiled compare in more places
nested models and deterministic models

richfitz:
This PR allows use of the compiled compare function in a few more places, using support added in https://github.com/mrc-ide/dust/pull/360

* deterministic models (was entirely unsupported)
* multistage models (required partially running the filter)

Merge https://github.com/mrc-ide/dust/pull/360 first and unpin branch
Fixes #177

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,7 +22,7 @@ BugReports: https://github.com/mrc-ide/mcstate/issues
 Imports:
     R6,
     callr (>= 3.7.0),
-    dust (>= 0.11.20),
+    dust (>= 0.11.23),
     processx,
     progress (>= 1.2.0)
 Suggests:
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -40,6 +40,6 @@ Suggests:
 RoxygenNote: 7.1.2
 Roxygen: list(markdown = TRUE)
 Remotes:
-    mrc-ide/dust@i177-partial-filter,
+    mrc-ide/dust,
     mrc-ide/odin.dust
 VignetteBuilder: knitr
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mcstate
 Title: Monte Carlo Methods for State Space Models
-Version: 0.8.3
+Version: 0.8.4
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("Marc", "Baguelin", role = "aut"),
@@ -40,6 +40,6 @@ Suggests:
 RoxygenNote: 7.1.2
 Roxygen: list(markdown = TRUE)
 Remotes:
-    mrc-ide/dust,
+    mrc-ide/dust@i177-partial-filter,
     mrc-ide/odin.dust
 VignetteBuilder: knitr
--- a/R/particle_filter_state.R
+++ b/R/particle_filter_state.R
@@ -147,12 +147,6 @@ particle_filter_state <- R6::R6Class(
 
       history <- self$history
       save_history <- !is.null(history)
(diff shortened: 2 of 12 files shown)


DESCRIPTION
```diff
@@ -40,6 +40,6 @@ Suggests:
 RoxygenNote: 7.1.2
 Roxygen: list(markdown = TRUE)
 Remotes:
-    mrc-ide/dust,
+    mrc-ide/dust@i177-partial-filter,
```
johnlees, line 43:
```suggestion
    mrc-ide/dust,
```

R/particle_filter_state.R
```diff
@@ -31,28 +31,134 @@ particle_filter_state <- R6::R6Class(
     min_log_likelihood = NULL,
     support = NULL,
 
-    run_compiled = function() {
+    step_r = function(step_index, partial = FALSE) {
```
johnlees, line 34: Similar to the comment on the dust PR, what's the reason for separating deterministic and stochastic, rather than having a single step function?

R/particle_filter_state.R
```diff
@@ -31,28 +31,134 @@ particle_filter_state <- R6::R6Class(
...
+      if (partial) {
+        log_likelihood_step
+      } else {
+        log_likelihood
       }
+    },
+
+    step_compiled = function(step_index, partial = FALSE) {
+      if (partial) {
+        stop("'partial' not supported with compiled compare")
```
johnlees, line 140: Maybe I'm confused, but I thought support was just added for this? Or am I misunderstanding what `partial` is?

tests/testthat/test-deterministic.R
```diff
@@ -362,3 +362,37 @@ test_that("Can offset the initial likelihood", {
...
+
+test_that("Can use compiled compare", {
+  dat <- example_sir()
+  p1 <- particle_deterministic$new(dat$data, dat$model, dat$compare, dat$index)
+  p2 <- particle_deterministic$new(dat$data, dat$model, NULL, dat$index)
+  pars <- list(exp_noise = Inf)
+  ## there's a discrepency here which ironically goes away with
+  ## non-infinite noise due to combination of sum over small numbers
+  ## and slightly different density calculation accuracy.
+  expect_equal(p1$run(pars), p2$run(pars), tolerance = 2e-5)
```
johnlees, line 375: Different example/non-Inf noise simpler then?

R/particle_filter_state.R
richfitz: partial is the support we put in for (one of) if2/smc2, where you get the likelihood contributed by the last single step. We don't support that but we do support running to a point that is not the end. It's confusing yes :)

richfitz: partial is the support we put in for (one of) if2/smc2, where you get the likelihood contributed by the last single step. We don't support that but we do support running to a point that is not the end. It's confusing yes :)

tests/testthat/test-deterministic.R
richfitz: Using an Inf noise keeps the calculation deterministic, even if not quite at the accuracy we'd like.  Getting the details out as to why this is slightly off would be potentially interesting but fairly sure that it's just that we have some poor accuracy in the densities when very unlikely (the R versions of these have lots of support to make these accurate but we can't use them due to thread safety)

R/particle_filter_state.R
richfitz: there's always a design trade-off here - one function with many if/elses or two functions with similar logic. Here the compiled and non-compiled versions are so different that the two functions approach feels much easier to follow the logic through. In particular the R step function is very complicated and has quite a few performance-critical tweaks

johnlees (approved):

PR #203, opened 2022-02-23
i187 rename pars_multi -> nested

### LINKED ISSUE #187: rename pars_multi
See https://github.com/mrc-ide/mcstate/pull/186#discussion_r783021822

Almost all internally facing except for if2 (on which nothing depends)

hillalex:
Fixes #187 
Although `pars_multi` is also a variable name in `dust` (and to change it there would be backwards incompatible as it is externally facing)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 .Rproj.user
 .Rhistory
 .RData
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mcstate
 Title: Monte Carlo Methods for State Space Models
-Version: 0.8.6
+Version: 0.8.7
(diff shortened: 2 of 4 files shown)


richfitz:
> Although pars_multi is also a variable name in dust (and to change it there would be backwards incompatible as it is externally facing)

I think that's ok. In dust, nested would not make a great deal of sense (though a better name could be found). There is really means that we have multiple sets of parameters, so one run of the filter will return multiple log likelihoods. For the uses in mcstate it's a specific use of this that leads to a model form that we're calling nested (as in hierarchical) with some shared and some separate parameters

richfitz (changes requested):
Thanks - this is great.

Could you increase the patch version number on this please (so to 0.8.7). For all these packages we update the drat repository (https://github.com/ncov-ic/drat) and it's useful to have the version number as installed by the user trace back to a single commit on master

richfitz (approved):

Issue #175, opened 2022-01-05
Check that pmcmc initial parameters are in range


richfitz:
* min, max and discrete all need checking, alongside the existing check of prior probability
* probably some duplication between simple and nested still to deal with

richfitz:
See the two comment blocks through similar code here: https://github.com/mrc-ide/mcstate/blob/1ace11aca59278a440fd3039f6fc51fa09051999/R/pmcmc.R#L126-L189

So, not where the user is defining the pmcmc_parameter object but where they provide both it and an initial condition, to ensure that the initial condition is correct

hillalex:
Oh those numbered TODOs are very useful, thanks :)

hillalex:
Although given an initial value is also provided when defining the params here:
https://github.com/mrc-ide/mcstate/blob/master/R/pmcmc_varied_parameter.R#L53 shouldn't it be checked to be within min/max here also?

richfitz:
Ah yes we should probably do the check too there, you're right. The user can override that initial value by the time it gets to the code linked above, so ideally we'd do it everywhere I think

PR #205, opened 2022-02-24
add 'integer' arg to params, deprecate 'discrete'

hillalex:
Ahead of adding parameter checks in #175, rename "discrete" to "integer", leaving the deprecated argument name in place for backwards compatibility.

### DIFF
--- a/R/if2_parameters.R
+++ b/R/if2_parameters.R
@@ -17,8 +17,12 @@
 ##'   `Inf`). If given, then `initial` must be at most this
 ##'   value.
 ##'
-##' @param discrete Logical, indicating if this parameter is
+##' @param discrete Deprecated. Logical, indicating if this parameter is
 ##'   discrete. If `TRUE` then the parameter will be rounded
+##'   after a new parameter is proposed.
+##'
+##' @param integer Logical, indicating if this parameter is
+##'   integer. If `TRUE` then the parameter will be rounded
 ##'   after a new parameter is proposed.
 ##'
 ##' @param prior A prior function (if not given an improper flat prior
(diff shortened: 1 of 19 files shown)


R/if2_parameters.R
```diff
@@ -17,10 +17,14 @@
 ##'   `Inf`). If given, then `initial` must be at most this
 ##'   value.
 ##'
-##' @param discrete Logical, indicating if this parameter is
+##' @param discrete Deprecated. Logical, indicating if this parameter is
 ##'   discrete. If `TRUE` then the parameter will be rounded
 ##'   after a new parameter is proposed.
```
richfitz, line 22:
```suggestion
##' @param discrete Deprecated; use `integer` instead.
```

R/pmcmc_parameters.R
```diff
@@ -17,10 +17,14 @@
 ##'   `Inf`). If given, then `initial` must be at most this
 ##'   value.
 ##'
-##' @param discrete Logical, indicating if this parameter is
+##' @param discrete Deprecated. Logical, indicating if this parameter is
```
richfitz, line 20: as above

R/pmcmc_parameters.R
```diff
@@ -30,9 +34,13 @@
 ##' @examples
 ##' pmcmc_parameter("a", 0.1)
 pmcmc_parameter <- function(name, initial, min = -Inf, max = Inf,
-                            discrete = FALSE, prior = NULL) {
+                            discrete, integer = FALSE, prior = NULL) {
+  if (!missing(discrete)) {
```
richfitz, line 38: This really highlights that there's some general parameter duplication we need to tidy up, probably before we get the ode stuff merged

richfitz (changes requested):
Thanks - can you bump the version number, and simplify the docs.

I've checked with Lilith and she's always passing `discrete = FALSE` in apparently...

R/pmcmc_parameters.R
hillalex: Same with the checking logic; the min/max/discrete checks need to run in a few places, so should be some way to share that code, whether a base parameter class (is that a thing in R?) or a shared function.

richfitz: Shared function probably easiest. You can do base classes with R6 but they're not used here

richfitz (approved):

PR #206, opened 2022-02-24
I175 check params

hillalex:


tests/testthat/test-pmcmc.R
```diff
@@ -291,7 +291,21 @@ test_that("can validate a vector of initial conditions", {
...
                "Expected names of 'initial' to match parameters ('a', 'b')",
                fixed = TRUE)
-  expect_error(pmcmc_check_initial(c(-0.1, 0.2), dat$pars, 1),
+  a <- pmcmc_parameter("a", 1, integer = TRUE)
+  pars <- pmcmc_parameters$new(list(a), diag(1) * 0.1)
+  expect_error(pmcmc_check_initial(1.1, pars, 1),
+               "'initial' must be an integer but was not (chain 1)",
+               fixed = TRUE)
+  expect_error(pmcmc_check_initial(c(-1, 1), dat$pars, 1),
+               "'initial' is less than 'min' (0, 0) (chain 1)",
```
hillalex, line 300: Not totally convinced by these messages, any suggestions for more useful phrasing here?

R/pmcmc.R
```diff
@@ -178,12 +164,38 @@ pmcmc_check_initial_nested <- function(initial, pars, n_chains) {
                      dimnames = list(nms, pops, NULL))
   }
 
-  ok <- apply(initial, 3, function(p) all(is.finite(pars$prior(p))))
+  pmcmc_check_initial_values(initial, pars, 3)
+}
+
+pmcmc_check_initial_values <- function(initial, pars, margin) {
```
richfitz, line 170: I think that margin here might be redundant as an arg - given the validation, is this always `length(dim(initial))`?

R/pmcmc.R
```diff
@@ -178,12 +164,38 @@ pmcmc_check_initial_nested <- function(initial, pars, n_chains) {
...
+  summary <- pars$summary()
+  ok <- apply(initial, margin,
+              function(p) all(round(p) == p | !summary$integer))
```
richfitz, line 173: I think we can subset with

```
array_first_dimension(initial, summary$integer)
```

then do the test against that more cleanly?

tests/testthat/test-pmcmc.R
richfitz: it's not clear here if the `0,0` refers to the expected min or the value given. Where things are guaranteed to be short we've thrown like:

```
value (X) is less than min (Y)
```

However, with these messages we could end up with really long strings if we're not careful - the sircovid model has ~60 parameters I think. It might be sufficient to report the name of the parameter that violates the constraint. It might also be worth detecting if _all_ chains fail the constraint, as that is likely in the default case and may simplify the reporting

R/pmcmc_parameters.R
```diff
@@ -46,6 +46,9 @@ pmcmc_parameter <- function(name, initial, min = -Inf, max = Inf,
   if (initial > max) {
     stop(sprintf("'initial' must be <= 'max' (%s)", max))
   }
+  if (integer & round(initial) != initial) {
+    stop(sprintf("'initial' must be an integer but was %s", initial))
+  }
```
richfitz, line 51: Can we route any of this through your new `pmcmc_check_initial_values` function?

richfitz (changes requested):

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: mcstate
 Title: Monte Carlo Methods for State Space Models
-Version: 0.9.0
```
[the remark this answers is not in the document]
hillalex: Ah yes this should probably be a patch as it isn't adding new functionality. But just to check, general rule is minor = new functionality, patch = bug fix?

tests/testthat/helper-mcstate.R
```diff
@@ -292,9 +321,11 @@ example_mvnorm <- function() {
 example_mvnorm_shared <- function(varied = TRUE, fixed = TRUE,
                                   proposal_varied = NULL,
                                   proposal_fixed = NULL) {
+
```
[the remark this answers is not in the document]
hillalex: Hmm, looks like it, and also looks like there is no way for me to turn this rule off

R/pmcmc_parameters.R
```diff
@@ -46,6 +46,9 @@ pmcmc_parameter <- function(name, initial, min = -Inf, max = Inf,
   if (initial > max) {
     stop(sprintf("'initial' must be <= 'max' (%s)", max))
   }
+  if (integer & round(initial) != initial) {
```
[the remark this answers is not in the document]
hillalex: Thanks. Presumably the `assert_scalar` call should actually be independent of the `integer` condition?

PR #209, opened 2022-05-11
Implemennt adaptive proposals for mcmc

richfitz:
Still needs some docs about how this works, and then tests of the inner workings once that is understood


MJomaba (approved):
Code looks good.

hillalex (approved):

PR #216, opened 2022-07-14
mrc-3406 expose mode control

hillalex:
Expose the stepper control for `mode` models via the particle filter interface.

### DIFF
--- a/R/particle_filter.R
+++ b/R/particle_filter.R
@@ -75,13 +75,15 @@ particle_filter <- R6::R6Class(
     ## Control for dust
     seed = NULL,
     n_threads = NULL,
+    ## Control for mode
+    control = NULL,
+    stochastic_schedule = NULL,
     ## Updated when the model is run
     last_stages = NULL,
     last_model = NULL,
     last_state = NULL,
     last_history = NULL,
-    last_restart_state = NULL,
-    stochastic_schedule = NULL
+    last_restart_state = NULL
   ),
 
   public = list(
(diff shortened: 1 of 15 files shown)


R/particle_filter.R
```diff
@@ -210,13 +212,16 @@ particle_filter <- R6::R6Class(
...
+    ##' @param control Tuning control for stepper
     initialize = function(data, model, n_particles, compare,
                           index = NULL, initial = NULL,
                           constant_log_likelihood = NULL,
                           n_threads = 1L, seed = NULL,
                           n_parameters = NULL,
                           gpu_config = NULL,
-                          stochastic_schedule = NULL) {
+                          stochastic_schedule = NULL,
+                          control = NULL) {
```
hillalex, line 224: this argument name is probably too vague. `stepper_control` better?

richfitz, line 224: Can we make this `ode_control` perhaps? (not `mode_control` though as we'll eventually merge these)

R/particle_filter.R
```diff
@@ -247,11 +252,18 @@ particle_filter <- R6::R6Class(
...
+          stop(paste("'stochastic_schedule' provided but 'model' does not",
+           "support this"))
+        }
+        if (!is.null(control)) {
+          stop(paste("'control' provided but 'model' does not",
+                     "support this"))
+        }
       } else {
         private$stochastic_schedule <- stochastic_schedule
+        private$control <- control
```
richfitz, line 266: Add

```
assert_is(control, "mode_control")
```

to validate that the control is ok here, and not when we initialise the mode model later.

(renaming control -> ode_control if you agree with the above)

richfitz (changes requested):
Small comments, nice how small this ends up being.

Can you also bump the version on this (and all mcstate PRs) - it'll fail on merge as it will try and tag the version, the version numbers are used by the downstream science users

R/particle_filter.R
richfitz: Sorry totally missed this comment here before :(

richfitz (approved):

PR #217, opened 2022-07-15
Return statistics from ode models

richfitz:
This PR exposes a `statistics()` method on the filter object, which returns statistics from the underlying model. Currently only valid for ode models and I am not sure that it would ever be useful on a discrete time model? We could call it `ode_statistics()` to reflect that or leave it generic and then if we change our mind it's there.

This does make me wonder if we should not be looking after the statistics for step in a similar way to state where we don't copy over - that's an issue for mode though as the interface won't change. But here we end up with the *filtered* set of steps which may not show us the particle that struggled a lot. Of course in the case with the malaria model they all struggle so it's probably ok

The second commit, 468a281, is not really related but fixes a bug while filling the remaining coverage holes (how can they hide in such small places?)

### DIFF
--- a/man/smc2_parameters.Rd
+++ b/man/smc2_parameters.Rd
@@ -13,18 +13,18 @@ used.
 \section{Methods}{
 \subsection{Public methods}{
 \itemize{
-\item \href{#method-new}{\code{smc2_parameters$new()}}
-\item \href{#method-sample}{\code{smc2_parameters$sample()}}
-\item \href{#method-names}{\code{smc2_parameters$names()}}
-\item \href{#method-summary}{\code{smc2_parameters$summary()}}
-\item \href{#method-prior}{\code{smc2_parameters$prior()}}
-\item \href{#method-propose}{\code{smc2_parameters$propose()}}
-\item \href{#method-model}{\code{smc2_parameters$model()}}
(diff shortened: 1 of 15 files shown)


man/smc2_parameters.Rd
```diff
@@ -13,18 +13,18 @@ used.
...
 \subsection{Public methods}{
 \itemize{
-\item \href{#method-new}{\code{smc2_parameters$new()}}
-\item \href{#method-sample}{\code{smc2_parameters$sample()}}
-\item \href{#method-names}{\code{smc2_parameters$names()}}
-\item \href{#method-summary}{\code{smc2_parameters$summary()}}
-\item \href{#method-prior}{\code{smc2_parameters$prior()}}
-\item \href{#method-propose}{\code{smc2_parameters$propose()}}
-\item \href{#method-model}{\code{smc2_parameters$model()}}
+\item \href{#method-smc2_parameters-new}{\code{smc2_parameters$new()}}
```
hillalex, line 16: looks like when I generated docs they came out in a different format, do you know why?

richfitz: You need to upgrade roxygen2 I think

```
> packageVersion("roxygen2")
[1] ‘7.2.0’
```

I think the last version update has a lot of churny change

hillalex (approved):

PR #219, opened 2022-08-05
Start work on nested adaptive fits

richfitz:


edknock:
This extends the adaptive proposal to use with the nested fits.

Also for both non-nested and nested:
- Adds output on the adaptive algorithm
- Adds a `diminishing_adaptation` option (`TRUE` as a default) to the adaptive proposal control, to provide more robust samples

PR #222, opened 2022-11-04
Update to use "time" not "step" in dust 0.12.x

richfitz:
Fairly robotic, but also fairly painful. Once this is passing there are a couple of rough edges that can be smoothed over as this package was already having a bit of a fight with both step and time being used

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mcstate
 Title: Monte Carlo Methods for State Space Models
-Version: 0.9.11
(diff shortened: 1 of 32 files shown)


richfitz:
This *was* read for review, until the that last version bump, which looks to have done something terrible to the version of dust installed; looks like remotes decided it would "upgrade" dust to a previous version on the logs. I'll open a separate PR that updates this to new actions, but in the meantime I think this is still ok to look at

hillalex (approved):

PR #223, opened 2022-11-18
Support for irregular time series

### LINKED ISSUE #97: support irregular steps
this is now trivial, with #96 merged, but changes are required to particle_filter_data, updating the docs and removing the now-unneeded checks. See 97e68ad and also the sircovid_data function which also simplifies

richfitz:
Fixes #97

### DIFF
--- a/R/particle_filter_data.R
+++ b/R/particle_filter_data.R
@@ -39,13 +39,15 @@
 ##'   integer-like for discrete time models and must be `NULL` for
 ##'   continuous time models.
 ##'
-##' @param initial_time Optionally, an initial time to start the model
-##'   from.  Provide this if you need to burn the model in, or if
-##'   there is a long period with no data at the beginning of the
-##'   simulation.  If provided, it must be a non-negative integer and
-##'   must be at most equal to the first value of the `time` column,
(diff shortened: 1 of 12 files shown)


R/particle_filter_data.R
```diff
@@ -117,37 +119,51 @@ particle_filter_data <- function(data, time, rate, initial_time = NULL,
...
-    if (initial_time > model_time_end[[1L]] - 1) {
-      stop(sprintf("'initial_time' must be <= %d", model_time_end[[1L]] - 1))
-    }
+  initial_time <- assert_integer(initial_time)
+  if (initial_time < 0) {
+    ## This condition is actually only required for discrete time
+    ## models; for continuous time models this would be fine.
+    stop("'initial_time' must be non-negative")
+  }
+  if (initial_time > model_time_end[[1L]]) {
```
hillalex, line 165: why was this check previously against `model_time_end[[1L] - 1`?

richfitz: I'm think this was that we used to require at least one time interval to have passed before the first data period, because we used to have the compare function take the current state and the previous state (so that cumulative variables could be converted into differences between reporting periods). But not entirely sure really, and it's certainly a complication that no longer seems required

hillalex (approved):

PR #235, opened 2023-09-04
Tidy format functions

lwhittles:
Adds a function converting the trajectory outputs of `pmcmc` into a tidy (long) format for easy (gg)plotting, with option to summarise over particles (outputting mean, median and 95% CrI as standard).

I have used the `tdigest::tdigest` function that allows very fast computation of quantiles, however this requires adding a new dependecy to the `tdigest` package. If this is not preferred we could just use the stats::quantile method in the package as standard and let users rewrite for efficiency if required (the sircovid use-case was pretty full-on most objects are unlikely to be as large)


R/pmcmc_tools.R
```diff
@@ -194,3 +194,100 @@ combine_state <- function(x) {
...
+}
+
+quantile_digest <- function(x, at) {
+
+  ## There's a garbage protection bug (possibly in
+  ## Rcpp) that is causing the tdigest object to get get collected. In
+  ## this case we fall back on quantile.
+  tryCatch(
+    tdigest::tquantile(tdigest::tdigest(x), at),
+    error = function(e) quantile(x, at, names = FALSE))
```
richfitz, line 260: actually, using base quantile but in a way that allows the user to bring their own might be nice. Or move tdigest to suggests and add a flag `tdigest` with a default of `FALSE`

R/pmcmc_tools.R
```diff
@@ -194,3 +194,100 @@ combine_state <- function(x) {
...
+
+pmcmc_tidy_chain_one <- function(object, type) {
+  assert_is(object, "mcstate_pmcmc")
+  stopifnot(type %in% c("pars", "probabilities"))
+
+  x <- object[[type]]
+  grid <- list(particle = seq_len(nrow(object$pars)),
+               type = type,
+               name = colnames(x))
+  if (object$nested) grid$population <- dimnames(x)[[3]]
```
richfitz, line 287:
```suggestion
  if (object$nested) {
    grid$population <- dimnames(x)[[3]]
  }
```

R/pmcmc_tools.R
```diff
@@ -194,3 +194,100 @@ combine_state <- function(x) {
...
+  ret <- do.call(expand.grid, grid)
+  ret$iteration <- object$iteration[ret$particle]
+  ret$chain <- object$chain[ret$particle] %||% 1
+  ret$value <- c(x)
```
richfitz, line 291:
```suggestion
  ret$value <- c(x) # drops unwanted attributes
```

richfitz (changes requested):
Thanks, this looks great. A couple of light suggestions below

R/pmcmc_tools.R
lwhittles: Like the idea of allowing a user to bring their own summary measure, might make sense to rename the resulting column `statistic` rather than `quantile` - what do you think?

R/pmcmc_tools.R
```diff
@@ -194,3 +194,108 @@ combine_state <- function(x) {
...
+##' @param summary_function Optional function to calculated user-defined
+##' summary statistics - defaults to calculating mean, 2.5-, 50-, and
+##' 97.5-percentiles
+##'
+##' @export
+##' @importFrom stats quantile
+
+pmcmc_tidy_trajectories <- function(object, summarise = FALSE,
+                                    quantiles = c(0.025, 0.5, 0.975),
+                                    summary_function = NULL) {
```
richfitz, line 223: Thinking about the interface here, perhaps:

````
pmcmc_tidy_trajectories <- function(object, summarise = FALSE, summary = NULL) {
  if (is.null(summary)) {
    summary <- function(x) {
      quantiles <- c(0.025, 0.5, 0.975)
      ret <- c(mean(x), quantile(x, quantiles, names = FALSE))
      set_names(ret, c("mean", paste0("q", quantiles * 100)))
    }
  }
```

then we document that this is the default. Otherwise we're holding a weird interface.

If the user does provide a function, it needs to take a single argument and return a named numeric vector

R/pmcmc_tools.R
```diff
@@ -194,3 +194,108 @@ combine_state <- function(x) {
...
+  if (object$nested) grid$population <- colnames(state)
+  if (summarise) {
+    # define summary_function and a vector of output names
+    f <- summary_function %||% pmcmc_summarise_one(quantiles)
+    summary_output <- f(numeric(1))
+    grid$statistic <- names(summary_output) %||% seq_along(summary_output)
+    } else {
+    grid$particle <- seq_len(nrow(object$pars))
+  }
+  grid$time <- object$trajectories$time
```
richfitz, line 242: Move this up above, and merge the two summarise if statements

R/pmcmc_tools.R
```diff
@@ -194,3 +194,108 @@ combine_state <- function(x) {
...
+  # the list entries will be dimnames, where they exist, and numeric otherwise
+  grid <- list(state = rownames(state) %||% seq_len(nrow(state)))
+  if (object$nested) grid$population <- colnames(state)
+  if (summarise) {
+    # define summary_function and a vector of output names
+    f <- summary_function %||% pmcmc_summarise_one(quantiles)
+    summary_output <- f(numeric(1))
+    grid$statistic <- names(summary_output) %||% seq_along(summary_output)
+    } else {
+    grid$particle <- seq_len(nrow(object$pars))
```
richfitz, line 240: I am a bit confused about what is going on here tbh. I think we should avoid the `f(numeric(1))` though and look at what was created in the subsequent block

richfitz (changes requested):
Sorry to send you around again on this. I can also just make this sort of change directly while holding your data structure constant if you prefer

